PR #10, opened 2017-12-22
Add React hot loader

nstrelow:
React hot loader refreshes parts of your React applications you changed without reloading the whole page or app. This enables fast iterations, extremely useful when adjusting the user interface.
Mostly followed this guid: https://github.com/gaearon/react-hot-loader
And used a similar structure as here: https://github.com/bpn1/curation

Essentially for hot reload to work, the AppRoot needs to be in a separate file and specified.
* Extracted Root of Application component to AppRoot.js
* Added necessary configurations to webpack.config and babel.js
* Used the HotReload AppContainer in index.js

### DIFF
--- a/frontend/.npmrc
+++ b/frontend/.npmrc
@@ -1,2 +1,0 @@
-registry=https://asfn7n79ca.execute-api.eu-west-1.amazonaws.com/prod/registry/
-always-auth=true


frontend/lib/js/store.js
```diff
@@ -0,0 +1,33 @@
+import {applyMiddleware, compose, createStore}   from "redux";
+import createHistory                             from 'history/createBrowserHistory';
+
+import reducers                                  from "./app/reducers";
+import {BASENAME, isProduction}                  from "./environment";
```
rs22: Same comment here

frontend/lib/js/index.js
```diff
@@ -4,24 +4,14 @@ import 'babel-polyfill';
...
 
 import './localization'; // To initialize locales
 import './app/actions'; //  To initialize parameterized actions
-
-import { BASENAME, isProduction } from './environment';
-import createMiddleware           from './middleware';
-
-import AppRouter                  from './app/AppRouter';
-import reducers                   from './app/reducers';
+import {store, browserHistory }        from './store'
```
rs22: Could you add some spaces between these curly braces?
Also, we typically use single quotes for includes, let's keep it consistent.

PR #23, opened 2018-01-18
Webpack upgrade & build optimizations

nstrelow:
Update webpack and include useful tools and optimizations
* Update all webpack dependencies to newest version
* Replace stats-webpack-plugin with WebpackBundleAnalyzer: Has same functionality, change `analyzerMode` to `server` to start a webapp to explore the size of dependencies
* Exclude unneeded moment locales as describe here: https://webpack.js.org/plugins/ignore-plugin/
* Add progress bar to webpack build for development
* Use `eval-source-map` as devtool. More info: https://webpack.js.org/configuration/devtool/


frontend/app/webpack.common.config.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
       inject: 'body',
       filename: 'index.html'
     }),
-    new webpack.NoEmitOnErrorsPlugin()
+    new webpack.NoEmitOnErrorsPlugin(),
+    new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/)
```
rs22: Looks like now that you've ignored all moment locales, the resulting strings are english (for example in the previous queries list). Can you have a look at it?

nstrelow: Yes, my bad. We should use the ContextReplacementPlugin instead: https://webpack.js.org/plugins/context-replacement-plugin/

Example:
```
new webpack.ContextReplacementPlugin(
  /moment[\/\\]locale$/, /de|fr|hu/
)
```

PR #28, opened 2018-01-18
English localizations and language-specific app bundles

### LINKED ISSUE #22: Default localization should be English
The default language for all the labels should be English. English should also be used in the demo screenshot on the front page

rs22:
This provides english localizations, along with some updates to how our bundling system produces language-dependent outputs.

The `main.js` file is now split up into a language-specific `main.en.js` or `main.de.js` file and a common `main.js` file (which is `require`d from both). In the process we're also setting the moment.js language accordingly.

Webpack is now configured to produce language-specific child-bundles, however we only generate one of them during development in order to reduce the initial compilation time during development.

When the express.js server is running in production mode, it chooses the respective localized version of the app based on the preferred language setting of the requesting browser.

Once the translations are approved, I will also create an updated screenshot, so this fixes #22.


frontend/app/src/js/main.de.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+import { initializeLocalization } from '../../../lib/js/localization';
+import translations               from '../../../lib/localization/de.yml';
+import conqueryTranslations       from '../localization/de.yml';
+
+initializeLocalization('de', translations, conqueryTranslations);
```
nstrelow: We should move the loading of `lib/localizations/[...].js` to the `initializeLocalization` to make the usage as library easier.
Even greater would be to only pass the locale string (e.g. 'de') and a function which loads localizations from app and lib and merges them.
This should be handled in another issue.

PR #35, opened 2018-01-19
Moving CI tag from travis.com to travis.org

manuel-hegner:


rs22:
The badge shows 'Build status unknown' right now because we didn't commit anything to the master since the travis switch.

Should we rather show the build status for develop (our 'default' branch on GitHub)?

nstrelow:
I agree, the status is mostly for us. Or we could simply display both master and develop.

manuel-hegner:
Why is develop even the default?

Nils Strelow <notifications@github.com> schrieb am Sa., 20. Jan. 2018,
17:06:

nstrelow:
Whether develop should be our main branch or not, I would simply display both. Preferably both with a heroku deployment to test it out.

manuel-hegner:
I would personally put the develop tags and Demo link into the wiki instead of the main page.

To master over develop:
- Of course we do. There should be a release on master every other week. People that want to get started developing with conquery should start with a stable build instead of a snapshot.
- Yes, because we do git-flow and thus there are rarely pull requests into master.
- That is true. But on the other hand a broken tag should also never be on develop, because pull requests can't be merged then. So maybe the badge is more of a identifiable link to the build system and in the future we will add badges about code health.
- That is acutally a good point and something I hate about github.

So I guess the default should stay at develop.

PR #44, opened 2018-02-02
Display version in development mode

### LINKED ISSUE #34: Frontend should show the current Version somewhere in the corner
THe Backend is already aware of its current version. This should be visible in the frontend somewhere (optimal solution would be to make this configurable).

nstrelow:
1. Loads the version from the /api/version route
2. Loads it into the redux store inside the header
3. Displays it inside the header
4. Is only displayed, when `isDevelopment` is true
5. Also displays version as title of logo

Closes #34


frontend/app/api/index.js
```diff
@@ -208,5 +208,12 @@ module.exports = function (app, port) {
...
       }, 500);
     }
-  )
+  );
+
+  /*
+    VERSION
+  */
+  app.get('/api/version', function(req, res) {
+    res.sendFile(path.join(__dirname, './version.json'))
```
rs22: Our demo backend is fake in a sense - but we might not have to fake this response. What about `require`ing the `package.json` and reading the npm package version from there? The `isDevelopment` value can also be easily obtained in node...

frontend/lib/js/header/Header.js
```diff
@@ -1,18 +1,46 @@
...
-import T     from 'i18n-react';
+import React           from 'react';
+import PropTypes       from 'prop-types';
+import T               from 'i18n-react';
+import { connect }     from 'react-redux';
+import { loadVersion } from './actions';
 
 class Header extends React.Component {
+  componentDidMount() {
+    this.props.loadVersion();
```
rs22: Oh, and for parsing URL parameters, I had designated the Startup component to be "the only" component in the app to dispatch actions in componentDidMount -- of course there may be exceptions to every rule, but I guess it helps to keep an overview about what's going on when the app starts.

What do you think? Could you integrate this call into that component?

frontend/app/api/index.js
nstrelow: @manuel-hegner wanted to send the value from the backend. In a production build we do not have the package.json but the version is still displayed as title tag of the conquery logo.

frontend/lib/js/header/Header.js
nstrelow: I thought about it, but I am unsure if this the right way.
It makes sense to only load the concepts at startup and only once to minimize traffic.

Isn't it better, when the data is not too big, to load it where it is needed? So that every component is connected to it's data and handles loading it.
The Startup component, it makes it difficult to understand where the data is coming from.

Let's say I am only looking at Header, then I need to find the reducer to version, the action and then the component executing that action (Startup).
Also now I don't need Header anymore and I delete that component. It is not obvious that one needs to adapt Startup and remove the API call there as it is not needed anymore and even if, you need to refactor code more locations than, when calling inside the component itself.

But this is of course if you do not know that everything is loaded in Startup.
Would you simply fetch the version in `componentDidMount` of the Startup component?

nstrelow:
It is the "alt text", actually title of the image. It is always displayed. I changed it to "Conquery {versionTag}"

frontend/app/api/index.js
rs22: This is demo backend code -- we sure have the package.json when running that backend in production!

nstrelow: Sorry, misunderstood you. You were completely right and I changed it using the package.json and NODE_ENV

PR #52, opened 2018-02-07
Add Travis selenium tests using Docker

MarcusBaitz:
This Configuration activated the Selenium tests in Docker over Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,14 @@
 language: java
 
+env:
+  - DOCKER_COMPOSE_VERSION=1.18.0
+
+before_install:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,5 +1,14 @@
 language: java
 
+env:
+  - DOCKER_COMPOSE_VERSION=1.18.0
+
+before_install:
+  - sudo rm /usr/local/bin/docker-compose
```
nstrelow: Why remove the local docker-compose?

MarcusBaitz: I get an error in the early version (see https://travis-ci.org/bakdata/conquery/builds/339475952)

PR #64, opened 2018-02-23
Use webpack-config-utils and optimize webpack

### LINKED ISSUE #29: Unify configs using webpack config utils
Right now we are using 3 different webpack config files. Production, Development and Common which is included in both. This is nice, but we need to call e.g. `...common.plugins` und plugins and all other nested objects.

With webpack config utils we would construct one single webpack config and provide alternatives in the code.

See https://doclets.io/kentcdodds/webpack-config-utils/master

An alternative would be to use webpack-merge and merge common and dev/prod.
See https://github.com/survivejs/webpack-merge

nstrelow:
Closes #29.

* Merged webpack configs into one, using [webpack-config-utils.](https://github.com/kentcdodds/webpack-config-utils) Now everything is in one place.
* Enabled scope hoisting, reduces size in production a bit (20kb). Detailed explanation [here](https://medium.com/webpack/brief-introduction-to-scope-hoisting-in-webpack-8435084c171f).
* Enabled babel-loader caching. Cached in .babel_loader_cache. [babel-loader options](https://github.com/babel/babel-loader#options)

I also suggest that we change devtool: 'eval-source-maps' to 'eval', if we do not use the generated js sourcemaps. This would speed up our development build.

An interesting article on optimizations [here](https://www.codementor.io/drewpowers/high-performance-webpack-config-for-front-end-delivery-90sqic1qa).
We could still investigate:
* Dynamic Imports: Imports and downloads some components only when needed. IMHO is not useful for us, only for bigger apps.
* CommonsChunkPlugin: Extract node_module packages to vendors. Reduces our main entry point and the file is cached in the browser as long as we don't upgrade our packages. User only needs to download our changed main.js

Webpack 4 should be out soon, this introduces nice default configurations for development and production.


nstrelow:
Closing in favor of webpack 4 PR: #128

PR #65, opened 2018-02-28
Markdown support for tooltip description added

MarcusBaitz:


rs22:
Where is this actually being used? Where in the app is a Tooltip's with a description set?

Could you maybe add some markdown to the demo data, so we can see it in action?

PR #70, opened 2018-03-02
Feature/deployment

MarcusBaitz:


frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "conquery",
-  "version": "1.2.0",
+  "name": "@bakdata/conquery",
```
rs22: Why did you decide on renaming the conquery npm package this way?

I've mostly seen orgs being used to group collections of packages from a specific 'domain', e.g.:
 - Babel plugins: https://www.npmjs.com/org/babel
 - TypeScript typings: https://www.npmjs.com/org/types

Adding an org-prefix for standalone projects adds some verbosity -- and as an open-source project I'd argue it's not necessary to include a company name in the package name.

MarcusBaitz: My concern was to organize the packages better in the future. Apparently I misunderstood the scope. I now know what you mean and change it. I understand that an organization is a group and not a company.

PR #72, opened 2018-03-07
Selenium Tests

MarcusBaitz:
This should be the basis for future Selenium tests.


selenium-test/src/main/java/com/bakdata/conquery/test/util/DragDropHelper.java
```diff
@@ -0,0 +1,302 @@
...
+            return 0;
+        }
+    }
+
+    private static String javaScriptEventSimulator = ""
+            + /* Creates a drag event */ "function createDragEvent(eventName, options)\r\n"
+            + "{\r\n"
+            + "var event = document.createEvent('HTMLEvents');\r\n"
+            + "event.initEvent('DragEvent', true, true);\r\n"
+            + //" var event = document.createEvent(\"DragEvent\");\r\n" +
```
rs22: Can this line be removed?

This is quite a lot of JavaScript code -- could this be loaded from a file to improve readability a bit?

PR #76, opened 2018-03-13
markdown support

### LINKED ISSUE #77: Markdown support
The frontend should support markdown notations in the additional info panel. We should also see if it would be feasible to support it in the descriptions of filters too.

MarcusBaitz:
- add [markdown](https://github.com/rexxars/react-markdown) support for additonal infos 
- on hide additional infos removed (mouse leave event)
- restore default html link color and text decoration

Do you mouse over on node Action Movies; see screenshot

selection_009

Closes #77


frontend/lib/styles/components/tooltip.sass
```diff
@@ -39,6 +39,9 @@
   &__key, &__value
     font-size: $font-xs
     margin: 0
+    a, a:visited, a:active
+      color: #0000EE
```
rs22: What do you think -- would `$col-blue-gray-dark` fit better into the overall design?

frontend/lib/js/tooltip/Tooltip.js
```diff
@@ -47,8 +48,8 @@ const Tooltip = (props: PropsType) => {
         {
           infos && infos.map((info, i) => (
             <div className="tooltip-info" key={i}>
-              <h3 className="tooltip-info__key">{info.key}</h3>
-              <p className="tooltip-info__value">{info.value}</p>
+              <Markdown className="tooltip-info__key" source={info.key} />
+              <Markdown className="tooltip-info__value"  source={info.value} />
```
rs22: I think there are two spaces before the source attribute...

frontend/app/api/concepts/movie_appearances.json
```diff
@@ -2,7 +2,13 @@
   "action_movies": {
     "parent": "movie_appearances",
     "label": "Action Movies",
-    "matchingEntries": 8234
+    "matchingEntries": 8234,
+    "additionalInfos": [
+      {
+          "key": "## Action film",
```
rs22: This will be rendered as an h2 -- however we apply some weird styling to h2 elements making them look like the headers of the editor tabs.

What should we do about it? We could limit that styling to the specific parts of the app, or should we discourage the use of markdown headers greater than, say, level-four headers?

PR #94, opened 2018-03-26
Release 1.8.0-RC1 back into develop

MarcusBaitz:


frontend/lib/js/form-components/ClearableInput.js
```diff
@@ -2,28 +2,53 @@
 
 import React                from 'react';
 import T                    from 'i18n-react';
+import NumberFormat         from 'react-number-format';
 
 import { isEmpty }          from '../common/helpers';
+import { MONEY_RANGE }      from '.';
```
rs22: Maybe a question of taste -- `import {} from '.'` is not very descriptive, we could also import this directly from `'./filterTypes'`

frontend/lib/js/form-components/InputRange.js
```diff
@@ -70,6 +84,11 @@ const InputRange = (props: PropsType) => {
           min: value ? value.min : null,
           max: value ? value.max : null,
           [type]: nextValue
+        },
+        {
+          min: props.input.formattedValue ? props.input.formattedValue.min : null,
+          max: props.input.formattedValue ? props.input.formattedValue.max : null,
+          [type]: formattedValue
```
rs22: I don't know why -- but when I was testing, the x-Buttons inside the input elements didn't have the expected effect. 

When resetting in exact mode, nothing happens, when in range-mode, only the formattedValue is reset but not the underlying raw value.

frontend/app/api/concepts.json
```diff
@@ -59,10 +59,20 @@
             {
               "id": "budget",
               "label": "Budget",
-              "type": "INTEGER_RANGE",
+              "type": "MONEY_RANGE",
               "unit": "1K USD",
```
rs22: This doesn't make sense when we're now entering Euro values.

frontend/lib/js/form-components/filterTypes.js
```diff
@@ -14,6 +15,7 @@ export const SUPPORTED_FILTERS = {
   REAL_RANGE: REAL_RANGE,
   STRING: STRING,
   BIG_MULTI_SELECT: BIG_MULTI_SELECT,
+  MONEY_RANGE: MONEY_RANGE,
   // Maybe later:
   // 'MONEY_RANGE'
```
rs22: Nitpick: Could you remove this comment line, now that we have a `MONEY_RANGE`? Could you also add `MONEY_RANGE` below `REAL_RANGE` above, so that all range filter types are grouped together?

frontend/lib/js/query-node-editor/ParameterTableFilters.js
```diff
@@ -120,7 +121,6 @@ const ParameterTableFilters = (props: PropsType) => (
                     inputType="number"
                     input={{
                       value: filter.value,
-                      defaultValue: filter.defaultValue,
```
rs22: Why did you remove this?

frontend/app/api/concepts.json
MarcusBaitz: This will render the correct filter type.

frontend/lib/js/query-node-editor/ParameterTableFilters.js
MarcusBaitz: I ask myself the same thing. I did not consciously remove that.

frontend/lib/js/form-components/InputRange.js
MarcusBaitz: Should be correct now

MarcusBaitz:
I had a branch on money range open, but by making a new delivery tomorrow, I've already merged the feature branch into the release branch.

frontend/app/api/concepts.json
rs22: I meant the `"unit": "1K USD"` line, I'll fix this online

frontend/.eslintrc.json
```diff
@@ -24,7 +24,7 @@
     "comma-dangle": 0, // May be there, helps better diffs and easier maintainability
     "indent": 0,                 // We indent how we like. (sometimes a custom indentation is great)
     "jsx-quotes": [2, "prefer-double"],
-    "max-len": [2, 100],
+    "max-len": [2, 120],
```
rs22: Was there a reason to change this?

frontend/app/api/concepts.json
```diff
@@ -59,10 +59,15 @@
...
               "description": "The movie's total budget",
-              "min": 1
+              "min": 1,
+              "defaultValue": {
```
rs22: I still think we should rather demo this in the example stored query, instead of always settings this default value. 
What movie has a budget smaller than 44,55 Euros? :wink:
Also, the value for `exact` has to be removed.

PR #96, opened 2018-03-26
New query node editor layout

### LINKED ISSUE #55: Initial dialog re-design
As a first step, I'll re-structure the exising modal dialog to follow the new design. This includes the following sub-tasks:

- [x] Implement the new design inside the QueryNodeModal
- [x] Provide an input box to edit concept list node titles
- [x] Make the QueryNodeModal self-contained to simplify its usage in multiple places in the codebase
- [x] Automatically create concept list nodes when dragging elements from the tree

rs22:
This PR finally introduces the UI changes for the Query Node Editor.

It defines three additional 'external' events for changing a concept list node's label, adding and removing concepts. These events are handled by the StandardQueryNodeEditor and the FormQueryNodeEditor implementations.

Furthermore, 'internal' state is definend (i.e. which tab is selected, which filter input is selected). However, consumers of the QueryNodeModal do not need to adopt to this, since the correct templated reducers were already registered.

Closes #55.


frontend/lib/js/query-node-editor/NodeDetailsView.js
```diff
@@ -0,0 +1,143 @@
...
+                'fa',
+                {
+                  'fa-square-o': !node.excludeTimestamps,
+                  'fa-check-square-o': node.excludeTimestamps
+                }
+              )} /> {T.translate('queryNodeEditor.excludeTimestamps')}
+            </button>
+          </div>
+        }
+
```
MarcusBaitz: the line could be removed

Issue #80, opened 2018-03-15
Create a search box at the top of the category tree view


rs22:
The search box is located under the dataset selection dropdown (analog to the previous queries search box). It allows to input any string and provides a button to clear its contents.

Issue #79, opened 2018-03-15
Implement filter function as part of a reducer


rs22:
The core search functionality will be provided by a reducer method that has the following features:

- [x] Listens for an action which will provide the user's search term as a payload
- [x] Traverses the concept trees, starting with the concepts referenced in the `categoryTrees` sub-state, retrieving concepts by their id using the `globalTreeStoreHelper`
- [x] Filters concepts using a simple string comparison on concept id, label, description and additional infos
- [x] Creates a flat array containing only the ids of the filtered concepts and their parents (without duplicates), i.e. all tree elements that should be displayed
- [x] Stores this array as part of the `categoryTrees` state object

All these features are covered by unit tests.

In order to access the `categoryTrees` state, the functionality has to be located near the `category-trees` reducer.

PR #103, opened 2018-04-09
Tree Search

MarcusBaitz:
- Create a search box at the top of the category tree view #80 
- Implement filter function as part of a reducer #79 
conquery


frontend/lib/js/category-trees/reducer.js
```diff
@@ -10,24 +10,113 @@ import {
...
+    function (match) { return '\\' + match }), "i");
+
+  return Object.assign({}, ...Object.entries(trees).map(([treeId, treeNode]) => ({
+    [treeId]: findTreeNodes(treeId, treeNode, regex)
+  })).filter(r => { return Object.keys(r).some(k => r[k].length > 0) }));
+}
+
+const findTreeNodes = (treeId: string, treeNode: NodeType, searchStr: string) => {
+  if (treeNode === null)
+    treeNode = getConceptById(treeId);
```
rs22: A couple of thoughts here:

 - Maybe we can avoid re-assigning method parameters
 - I've had a quick look at the CPU profiler: getConceptById takes up about 70% of the total search time. I'm not sure if we can improve on that in general, but the specific implementation doesn't seem optimal. Maybe we can have a variant of that method where we pass in the concept tree id as a parameter, rather than letting the method itself search all trees for the provided id.
 - Let's be consistent with naming (here and in general): When something is called `tree` or `treeId` it should always refer to the root node of a concept tree (i.e. the one with `tables` set, **not** a structure node and **not** any other node of a concept tree)

frontend/lib/js/category-trees/reducer.js
```diff
@@ -10,24 +10,113 @@ import {
...
+
+  const children = treeNode.children || [];
+  const result = children.map(child => findTreeNodes(child, null, searchStr))
+    .reduce((agg, cur) => [...agg, ...cur], []);
```
rs22: One idea for optimization (don't know if that's solving our problem): Perform `findTreeNodes` inside the `reduce` call and stop once a certain number of results was found.

frontend/lib/js/category-trees/reducer.js
```diff
@@ -10,24 +10,113 @@ import {
...
+  const description = treeNode.description || '';
+  const additionalInfos = treeNode.additionalInfos
+    ? treeNode.additionalInfos.map(t => { return t.key + " " + t.value   }).join('')
+    : '';
+
+  if (result.length ||
+      label.match(searchStr) ||
+      description.match(searchStr) ||
+      additionalInfos.match(searchStr))
+        return [treeId, ...result];
```
rs22: Indentation (two spaces more than `if`)

frontend/lib/js/category-trees/reducer.js
```diff
@@ -10,24 +10,113 @@ import {
...
+
+  return [];
+}
+
+const fuzzyMatch = (text: string, searchStr: string) => {
```
rs22: This method isn't used (and the linter should have complained about it)

frontend/lib/js/form-components/SearchBox.js
```diff
@@ -2,38 +2,50 @@ import React                           from 'react';
 import PropTypes                       from 'prop-types';
 import T                               from 'i18n-react';
 import { Creatable as Select }         from 'react-select';
+import { isEmpty } from '../common/helpers';
```
rs22: Alignment

frontend/lib/js/form-components/SearchBox.js
```diff
@@ -2,38 +2,50 @@ import React                           from 'react';
...
+            { label }
+          )}
+          placeholder={T.translate('reactSelect.searchPlaceholder')}
+          backspaceToRemoveMessage={T.translate('reactSelect.backspaceToRemove')}
+          clearAllText={T.translate('reactSelect.clearAll')}
+          clearValueText={T.translate('reactSelect.clearValue')}
+          noResultsText={T.translate('reactSelect.noResults')}
+        />
+      </div>
+    : props.mode === 'simple' &&
```
rs22: You could've made mode a boolean property, that allows for a nice React syntax, i.e. `<SearchBox isMulti />`

frontend/lib/js/form-components/SearchBox.js
```diff
@@ -2,38 +2,50 @@ import React                           from 'react';
...
+      <div className="search-box input--full-width">
+        <ClearableInput
+          inputType="text"
+          placeholder={T.translate('reactSelect.searchPlaceholder')}
```
rs22: The placeholder text doesn't look the same compared to the previous queries search box...

frontend/lib/js/category-trees/reducer.js
```diff
@@ -10,24 +10,113 @@ import {
...
+    treeNode = getConceptById(treeId);
+
+  const children = treeNode.children || [];
+  const result = children.map(child => findTreeNodes(child, null, searchStr))
+    .reduce((agg, cur) => [...agg, ...cur], []);
+
+  const label = treeNode.label || '';
+  const description = treeNode.description || '';
+  const additionalInfos = treeNode.additionalInfos
+    ? treeNode.additionalInfos.map(t => { return t.key + " " + t.value   }).join('')
```
rs22: This can be `treeNode.additionalInfos.map(t => t.key + " " + t.value)`, or even ``treeNode.additionalInfos.map(({key, value}) => `${key} ${value}`)``.

Shouldn't we be doing `.join(' ')`?

frontend/lib/js/category-trees/reducer.js
```diff
@@ -10,24 +10,113 @@ import {
...
+      additionalInfos.match(searchStr))
+        return [treeId, ...result];
+
+  return [];
+}
+
+const fuzzyMatch = (text: string, searchStr: string) => {
+    if (!text) return '';
+
+    var search = searchStr.replace(/ /g, '').toLowerCase();
```
rs22: Let's not use `var` anymore.

frontend/lib/js/category-trees/CategoryTree.js
```diff
@@ -10,6 +10,7 @@ import type {
 }                           from '../common/types/backend';
 
 import CategoryTreeNode     from './CategoryTreeNode';
+import { SearchType }       from './reducer';
```
rs22: `type` missing in import

frontend/lib/js/category-trees/CategoryTreeNodeTextContainer.js
```diff
@@ -50,17 +54,30 @@ const CategoryTreeNodeTextContainer = (props: PropsType) => {
           <i className={classnames(
             'category-tree-node__icon',
             'fa', {
-              'fa-folder-open': !!props.open,
-              'fa-folder': !props.open
+              'fa-folder-open': !!props.open || searching,
+              'fa-folder': !props.open || !searching
```
rs22: Consider the case `props.open == false` and `searching == true`. This will result in both classes being applied. So we need `!props.open && !searching`

frontend/lib/js/category-trees/actions.js
```diff
@@ -90,3 +92,8 @@ export const loadTree = (datasetId: DatasetIdType, treeId: TreeNodeIdType) => {
       );
   };
 };
+
+export const searchTreesStart = (searchStr: string) =>
+  ({type: SEARCH_TREES_START, payload: { searchStr }});
+export const searchTreesEnd = (searchStr: string, result: any) =>
```
rs22: This isn't used anywhere (yet), is it? 

Maybe you should move the search result computation into a redux thunk (although there's nothing asynchronous going on at the moment). 

It's nice that we already have the prerequisites for computing the search results on the backend, but we should make sure that we don't forget about this action type by using it somewhere.

frontend/lib/js/category-trees/CategoryTreeSearchBox.js
```diff
@@ -0,0 +1,14 @@
+import { connect } from 'react-redux';
```
rs22: `// @flow` is missing

frontend/.flowconfig
```diff
@@ -17,4 +17,4 @@ module.name_mapper='.*\(.yml\)' -> 'empty/object'
 ; suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
 
 [version]
-0.66.0
+0.65.0
```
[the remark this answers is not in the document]
MarcusBaitz: Since the conversion, I no longer have memory leak in vscode/flow. I'll adapt it in the package.json or I only work locally with this version?

frontend/lib/js/category-trees/CategoryTreeFolder.js
```diff
@@ -21,6 +25,7 @@ type PropsType = {
   active: boolean,
   open?: boolean,
   onToggleOpen?: Function,
+  search: SearchType,
```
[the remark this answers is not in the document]
MarcusBaitz: Because the `search` is an integral part of the app. In the beginning it was optional. Correctly, it should be optional. I change it.

i.e. `open` is not always optional.

frontend/lib/js/category-trees/CategoryTreeSearchBox.js
```diff
@@ -0,0 +1,14 @@
...
+import { SearchBox } from '../form-components';
+import { searchTreesStart } from './actions';
+
+const mapStateToProps = (state) => ({
+  searchStr: state.categoryTrees.search.searchStr,
```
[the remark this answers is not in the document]
MarcusBaitz: `query` is already exists. I think something out.

frontend/package.json
```diff
@@ -97,6 +99,7 @@
     "html-webpack-plugin": "^2.30.1",
     "husky": "^0.14.3",
     "ignore-styles": "^5.0.1",
+    "js-worker-search": "^1.2.1",
```
rs22: Do we still need this?

frontend/lib/js/form-components/SearchBox.js
```diff
@@ -1,39 +1,84 @@
...
+          searchResult.loading
+          ? <span className="dots"><Dot>.</Dot><Dot>.</Dot><Dot>.</Dot></span>
+          : searchResult.searching && searchResult.resultCount >= 0 &&
+            <span className="input input-label--disabled input-label--tiny">
+              {
+                T.translate('search.resultLabel', {
+                  limit: searchResult.limit,
+                  resultCount: searchResult.resultCount,
+                  duration: duration(
+                      searchResult.duration,
```
rs22: Seems to be indented a little too far...

frontend/lib/js/category-trees/selectors.js
```diff
@@ -0,0 +1,33 @@
...
+      }
+    }
+    return false;
+}
+
+export const isInSearchResult = (id: TreeNodeIdType, children?: [], search?: SearchType) => {
+    if (!search || !search.result) return false;
+    const result = search.result;
+
+    for (var i = 0; i < result.length; i++) {
```
rs22: You could also write `if (result.some(resultId => resultId.indexOf(id) >= 0)) { ... }`, right?

frontend/lib/js/category-trees/selectors.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+import { type TreeNodeIdType }   from "../common/types/backend";
+import { type SearchType }       from "./reducer";
```
rs22: Being picky again... I'd vote for using single quotes here

frontend/lib/js/category-trees/reducer.js
```diff
@@ -10,24 +10,93 @@ import {
...
 };
 
+const setSearchTreesEnd = (state: StateType, action: Object): StateType => {
+  const { query, result } = action.payload;
+  const searching = query && query.length > 0;
+  const maxResult = [];
+  const limit = parseInt(process.env.SEARCH_RESULT_LIMIT);
+
+  for (var i = 0; i < result.length; i++) {
+    maxResult.push(result[i]);
```
rs22: Wouldn't it make sense to only return so many results from the server? We could add a query parameter to the search request to send the current `SEARCH_RESULT_LIMIT`. Then, the server would need to supply the number of total results as wel...

frontend/lib/js/category-trees/actions.js
```diff
@@ -90,3 +93,22 @@ export const loadTree = (datasetId: DatasetIdType, treeId: TreeNodeIdType) => {
...
+export const searchTreesStart = (query: string) =>
+  ({type: SEARCH_TREES_START, payload: { query }});
+export const searchTreesEnd = (query: string, result: []) =>
+  ({type: SEARCH_TREES_END, payload: { query, result }});
+export const searchTreesError = (query: string, err: any) =>
+  defaultError(SEARCH_TREES_ERROR, err, { query });
+
+export const searchTrees = (datasetId: DatasetIdType, query: string) => {
+  return (dispatch: Dispatch) => {
+    dispatch(searchTreesStart(query))
```
rs22: We should check somewhere if the query is empty (i.e. has been cleared) -- in this case a server request results in an error, and we have a short delay before the unfiltered tree shows.

frontend/lib/js/category-trees/selectors.js
MarcusBaitz: No, it is a performance aspect. i've record the performance: the method **some** was under the top slowest in this implementation.

frontend/lib/js/category-trees/selectors.js
MarcusBaitz: right.

frontend/lib/js/category-trees/reducer.js
MarcusBaitz: yes, agree. That was also my consideration but then we need a new display of the result set or this can be removed.

frontend/lib/js/category-trees/actions.js
MarcusBaitz: The server request is now not executed on an empty string

PR #128, opened 2018-07-11
Upgrade to webpack 4 and unify webpack configs

nstrelow:
Webpack
* __upgrade to webpack 4, comes with many default optimizations for dev and prod__
* __Use [webpack-config-utils](https://github.com/kentcdodds/webpack-config-utils) to unify webpack configs__
* Use [MiniCssExtractPlugin ](https://github.com/webpack-contrib/mini-css-extract-plugin) instead of ExtractTextPlugin (used for CSS only and highly recommended by ExtractTextPlugin for webpack 4)
* Use one rule for sass and css -> ExtractText/MiniCssExtract now also extracts .css files
* Remove indentedSyntax, is now automatically set per file (needed for sass/css): https://github.com/webpack-contrib/sass-loader/pull/196
* Rule for .yml: Remove depracated json-loader and use [js-yaml-loader](https://github.com/wwilsman/js-yaml-loader) to directly convert to JS
* Remove DefinePlugin process.node.ENV as this is set now automatically by mode
* Use OptimizeCSSAssetsPlugin to minify CSS
* Parallelize for UglifyJS. (Without optimize.minify, webpack would default to UglifyJS in production mode without parallelization)

Other
* Use const instead of var in server.js


nstrelow:
PS: Wasted a few days upgrading to webpack 4 and cleaning up the webpack config. I hope I never have to do that again.

Kadrian:
I do like this in general, but I'd have to test it before merging. So would vote to keep this open

nstrelow:
I would suggest using this as an example to build a new webpack config. Not all packages need to be upgraded, but I would upgrade the version numbers of the webpack ones

Kadrian:
Closing this in favor of #335

PR #129, opened 2018-07-11
DropzoneList exposes a flag to restrict number of dropzones

m-wettig:
* add the optional flag `disallowMultipleColumns`
* if absent or false DropzoneList will render any number of additional dropzones
* if true only the initial dropzone will be rendered. Then, concepts can only be added to the very first group


m-wettig:
@MarcusBaitz I need this change for v1.9

PR #231, opened 2019-01-04
Fixed not having a dedicated FilterValue for BigDecimal Ranges.

martintaraz:
import from https://github.com/bakdata/conquery-backend/pull/200


awildturtok:
Sum_Integer fails when run in batch, but succeeds when run alone? @manuel-hegner

PR #237, opened 2019-01-07
Temporal Queries

manuel-hegner:
import from https://github.com/bakdata/conquery-backend/pull/97


awildturtok:
DaysBeforeOrNever fails when run in batch mode, but doesn't when run alone. I've checked the output values and they are correct so there's something else at play here. @manuel-hegner

awildturtok:
@manuel-hegner: I would request review if you weren't the owner of this PR

awildturtok:
🎉🎉🎉

PR #286, opened 2019-02-04
Query Tests ICD mit 1. KH-Tabelle und 2. KH und AU-Tabelle

ana-suput:
1. Jahr 2017, ICD F20, KH-Tabelle
2. Jahr 2017, ICD F20, KH- und AU-Tabelle


manuel-hegner:
@awildturtok This is an error in the ConceptTreeCache. The problem occurs here because the concept has multiple connectors and the current cache implementation is just caching based on the int valueId. But a valueId is only comparable to other valueIds within the same Import. The cache should require the import with the valueId (see call in CalculateCBlocksJob#97) and cache in a unique domain for each import. Remember that the cache needs to be thread-safe.

PR #291, opened 2019-02-06
Aggregator/Filter composition via Ids.

awildturtok:


awildturtok:
@manuel-hegner Please review this.

awildturtok:
This branch is intentionally left to stale, as it will create a lot of friction with the existing codebase. This branch will be merged when development resumes to a less critical period.

thoniTUB:
@awildturtok is this still relevant?

awildturtok:
This is a proposal which would effectively rewrite the entire QueryEngine.

Filters are internally just wrappers around aggregators which is why we could implement them exactly as such in the Query Layer. This would allow more expressiveness in the queries as well as reducing redundancy: At the moment, if a user uses a select and filters on the corresponding filter, the aggregator will be executed twice.

PR #335, opened 2019-02-12
Feature/upgrade to webpack 4

Kadrian:
This upgrades

- `webpack`
- `babel`
- `flow`

and uses `jest` instead of `mocha`, `chai` and `nyc`.

This PR is inspired by and replaces #128 

**Minor**

- Removed code coverage for now, since we're *really* not covering much yet.

## Things I checked

- App still starts (`yarn start`) and seems to run in browser without errors
- App builds (`yarn run build-app`) in both `en` and `de`.
- I made sure that this funny `gulp` setup (`yarn build`) still works. Had to upgrade `gulp-babel` as well.
- Tests run (`yarn test`)


Kadrian:
Rebased on #336

PR #379, opened 2019-02-23
Disentangle standard query editor filter setting logic

Kadrian:
This removes the entire "options" handling block in `setNodeFilterProperties` and adjusts the callees.

Also removes the helper method `stripObject` which made everything more complicated as it set properties explicitly to `undefined`. That's an antipattern we'll migrate away from.

This PR has several benefits:

- Fixes the error that occurred when entering anything into a range (`options` was `null` in that case)
- Probably fixes the hanging application when a long filter list is dropped. But that has to be tested further.
- Logic becomes simpler to reason about: `setNodeFilterProperties` simply accepts an object and sets it as a filter. Calling methods have to take care about sending the right properties.

**Note: Apologies that it's not easy to review – I recently introduced prettier and now the first time I touch a file, there are a lot of formatting fixes**


manuel-hegner:
Please use feature/XYZ for future branches. The feature directory is here not meant as a semantic tag but we follow gitflow which groups everything except master/develop under feature.

PR #396, opened 2019-02-27
validity date refactoring

manuel-hegner:
validitydateselection is now its own thing


manuel-hegner:
With the introduction of selects I want to refactor the validity date selection Before now this was just modeled as a filter so the frontend required no specific handling.

With this PR the table json object has a new field `validityDateSelection` that contains the possible values. The frontend should render this as a single value select where the first given value is already selected and there is no possibility to select no value (this is already supported).

The result should be written in the resulting json object as `validityDate`.

Kadrian:
I agree that this is a good idea, since it's a very important kind of filter.

Now since we have `tables`, `selects`, `filters` and similar wording already, I would like to propose calling this `dateColumn` instead of `validityDate`.

To keep the structure similar to the `filters`, and to be able to use a `defaultValue` I'd like to propose the following.

Table object from the API
``` 
{
  ...
  dateColumn: {
    defaultValue: "ABC",
    options: [
      { value: "ABC", label: "Super date" },
      { value: "EFG", label: "Boring date" },
    ]
  }
}
```

**Table object sent to the API**
```
{
  ...
  dateColumn: {
    value: "ABC"
  }
}
```

manuel-hegner:
Mh, seems reasonable, but can we use `validityDateColumn`? dateColumn alone is not expressing wuite the idea behind this selection.

Kadrian:
=> Haben uns nach reichlicher Diskussion auf `dateColumn` geeinigt :-)

manuel-hegner:
backend part is done

Kadrian:
Frontend part is done as well. Here, like discussed:

Bildschirmfoto 2019-07-05 um 17 40 29

PR #406, opened 2019-03-01
Querytests with empty dates

ana-suput:
SUM and DURATION_SUM querytest to make sure that data with empty validity-dates is not included if there is a date restriction


manuel-hegner:
@ana-suput bitte up-to-date bringen

PR #435, opened 2019-03-11
fix and test for decimal type and appropriate filter.

awildturtok:


backend/src/main/java/com/bakdata/conquery/models/types/specific/DecimalTypeScaled.java
```diff
@@ -38,7 +39,7 @@ public NUMBER transformFromMajorType(DecimalType majorType, Object value) {
 		BigDecimal v = (BigDecimal) value;
 		if(v.scale() > scale)
 			throw new IllegalArgumentException(value+" is out of range");
-		return subType.transformFromMajorType(null, v.unscaledValue().longValueExact());
+		return subType.transformFromMajorType(null, v.setScale(scale, RoundingMode.HALF_DOWN).unscaledValue().longValue());
```
manuel-hegner: RoundingMode.ROUND_UNNECESSARY
 wouldn't that be better here since rounding should never be necessary?

PR #437, opened 2019-03-11
proposing new directory structure to bring yarn and maven more in lin…

manuel-hegner:
…e with each other


Kadrian:
Looks good to me except the change in `frontend/app/api/dnd/filter-values-many.txt` (which I'm not getting. Is that a line endings thing?)

PR #439, opened 2019-03-11
Removed Filters out of FilterNodes

awildturtok:
They were only referenced for columns and could therefor be refactored into a column reference.


awildturtok:
@manuel-hegner please review

PR #509, opened 2019-04-10
experimenting with autoformatting

manuel-hegner:


manuel-hegner:
@awildturtok @martintaraz @thoniTUB 
You think this is a useful thing to do or to restrictive and annoying durin normal coding?
(to explain: in this branch I added a travis job that autoformats the code and pushes resulting changes after every push)

awildturtok:
Personally, I would prefer it if PRs are marked as non compliant and the autoformatting is done client side, on save (or as pre/post commit hook, though i'm not a huge fan of that).

I anticipate some very strange merge (conflicts) happening when someone merges a branch they have been working on offline for a while or simply forgot to push/maintain. Also it is imho a little too intrusive in the git history. I'd much prefer if I see what I will be publishing.

Kadrian:
In the frontend, with prettier.js we're also formatting on the developer's side. Basically, developers are required to install auto formatting in their IDEs / development process. I think that's a fair trade off, because I'd also argue that conflicts might arise or unexpected fixes might become necessary later on, when formatting happens after a push.

PR #510, opened 2019-04-11
adds a method to collect all namespaced Ids in a CQuery

thoniTUB:
Adds a method to retrieve all namespacedIDs in a concept query, this is necessary for an external service


thoniTUB:
@manuel-hegner should i ignore the style issues or add classes like CQChainElement and CQConjuctionElement to avoid code duplication

PR #516, opened 2019-04-15
Perform concept search entirely in frontend

Kadrian:
This moves searching the concepts into the frontend.

**Why does this work?**
In general, this works because all concepts are lying in the frontend anyways.

It works **in particular**, because the results remain folded with result number indications. That means that not too much has to be rendered directly after searching.

And it's also quite fast.

- I've tested it with roughly ~150,000 concepts and the search is almost instant for me (~0.3s).
- Even with ~80000 results when searching, e.g: "an", the search doesn't take longer than 0.8s.
- Throttling my CPU by 6x, even the longest search didn't take longer than 2.8s.

**Why is it so fast?**
I've made sure the search algorithm has linear time complexity (O(n)) and avoids costly Javascript features, such as object spread.

**Why is that so cool?**
1) Backend search can be removed @awildturtok @thoniTUB. We won't need to maintain the search full-stack anymore.
2) We save a little traffic (we've been pushing ~50k strings through the wire)
3) We save costly backend calculations, which directly translates to: saving hosting costs :-)

**And what's the coolest thing about all of that?**
Bildschirmfoto 2019-04-15 um 21 43 31

### DIFF
--- a/frontend/lib/js/category-trees/globalTreeStoreHelper.js
+++ b/frontend/lib/js/category-trees/globalTreeStoreHelper.js
@@ -137,42 +137,6 @@ export const search = async (query: string) => {
   });
 };
 
-const findConcepts = (treeId, nodeId, node, query, intermediateResult) => {
-  const isNodeIncluded = doesQueryMatchNode(node, query);
-
-  let sum = isNodeIncluded ? 1 : 0;
-
-  if (node.children) {
-    for (let child of node.children) {
-      const result = findConcepts(
-        treeId,
-        child,
-        window.categoryTrees[treeId][child],
-        query,
-        intermediateResult
-      );
-
-      sum += result[child] || 0;
-    }
-  } else {
-    if (isNodeIncluded) {
-      intermediateResult[nodeId] = 1;
-
-      return intermediateResult;
-    } else {
-      return intermediateResult;
-    }
-  }
-
-  if (sum === 0) {
-    return intermediateResult;
-  } else {
-    intermediateResult[nodeId] = sum;
-
-    return intermediateResult;
-  }
-};
-
 const doesQueryMatchNode = (node, query) => {
   const lowerQuery = query.toLowerCase();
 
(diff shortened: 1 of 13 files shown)


frontend/lib/js/category-trees/globalTreeStoreHelper.js
```diff
@@ -120,3 +120,95 @@ export const hasConceptChildren = node => {
...
+    {}
+  );
+
+  return result;
+};
+
+const doesQueryMatchNode = (node, query) => {
+  const lowerQuery = query.toLowerCase();
+
+  return (
```
awildturtok: Du erzeugst hier viele kurzlebige objekte, wäre es nicht sinnvoller die irgendwie zu cachen oä? Oder zumindest aufbereitet vorzulegen?

Kadrian:
Even tested in IE11 and Edge, seems to work!

frontend/lib/js/category-trees/globalTreeStoreHelper.js
Kadrian: Welche Objekte meinst du? 

Wenn ich jetzt raten müsste: Theoretisch könnte man die "geLowerCased" `label` / `description` und die "gejointen" `additionalInfos` vorher berechnen / cachen. Weil die Suche aber sowieso so schnell ist, ist das denke ich nicht notwendig.

Kadrian: Um noch konkreter zu sein: Hier ein Benchmark, Javascript ist relativ zügig bei den Operationen. Selbst beim langsameren join + lowercase Beispiel schafft mein Macbook 2,6 mio Operationen / Sekunde. Wir können also selbst auf einem 10x langsameren Rechner > 200k Ergebnisse in unter einer Sekunde durchsuchen. Würde sagen, das passt erstmal.

Bildschirmfoto 2019-04-16 um 11 05 34

PR #530, opened 2019-04-24
Pass an array of children to concept_query in case of an external query

Kadrian:
**My goal**

Detect and external query and don't allow to "expand it".

**Current status**
Right now, in case of an external query (uploading a query result), the frontend sends a single element as children, like:

```
{
  type: "CONCEPT_QUERY",
  root: {
    type: "AND",
    children: {
      type: "EXTERNAL",
      ...
    }
  }
}
```

Somehow, the backend seems to be able to deal with that, but here, I'm suggesting to change this into:
```
{
  type: "CONCEPT_QUERY",
  root: {
    type: "AND",
    children: [{
      type: "EXTERNAL"
      ...
    }]
  }
}
```

But really, it doesn't seem to be correct to wrap this in `CONCEPT_QUERY` and `AND` in the first place. 

@manuel-hegner @awildturtok What do you think?

### DIFF
--- a/frontend/lib/js/api/apiHelper.js
+++ b/frontend/lib/js/api/apiHelper.js
@@ -171,7 +171,7 @@ const transformTimebasedQueryToApi = query => ({
 });
 
 const transformExternalQueryToApi = query =>
-  createConceptQuery(createExternal(query));
+  createConceptQuery([createExternal(query)]);
 
 const createExternal = (query: any) => {
   return {
--- a/frontend/lib/js/api/apiHelper.js
+++ b/frontend/lib/js/api/apiHelper.js
@@ -75,14 +75,16 @@ export const transformElementsToApi = conceptGroup =>
   conceptGroup.map(createConcept);
 
 const transformStandardQueryToApi = query =>
-  createConceptQuery(createQueryConcepts(query));
+  createConceptQuery(createAnd(createQueryConcepts(query)));
 
-const createConceptQuery = children => ({
+const createConceptQuery = root => ({
   type: "CONCEPT_QUERY",
-  root: {
-    type: "AND",
-    children: children
-  }
+  root
+});
+
+const createAnd = children => ({
+  type: "AND",
+  children
 });
 
 const createNegation = group => ({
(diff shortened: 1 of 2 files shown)


Kadrian:
Also, the problem is that currently, an external query seems to be resolved like this:

```
{
  "query": {
    "type": "CONCEPT_QUERY",
    "root": {
      "type": "AND",
      "children": [
        {
          "type": "OR",
          "children": [
            {
              "type": "SAVED_QUERY",
              "query": "dataset.$370bd80eb$2d40ea$2d4f7a$2daac4$2de310e3187ee9",
              "resolvedQuery": {
                "type": "CONCEPT_QUERY",
                "root": {
                  "type": "AND",
                  "children": [
                    {
                      "type": "EXTERNAL_RESOLVED",
                      "values": {
                        "-1": [
                          16323,
                          16323
                        ]
                      }
                    }
                  ]
                }
              }
            }
          ]
        }
      ]
    }
  }
}
```

which is a little *complicated*. How can we improve this?

manuel-hegner:
mhhh
The resolved part is actually only used internally. It is an error that that is send to the frontend.
That the EXTERNAL nodes are wrapped in AND and OR nodes is only the case because the frontend send them like this.
That the array is missing in the json is because we accept single elements innstead of arrays. It seems that also means we write single element arrays as single elements, which we should not.

Kadrian:
You're right, the resolved query I pasted is actually an external query re-used inside of a query. So the actual part here is sent back to the frontend
```
{
  "type": "CONCEPT_QUERY",
  "root": {
    "type": "AND",
    "children": [
      {
        "type": "EXTERNAL_RESOLVED",
        "values": {
          "-1": [
            16323,
            16323
          ]
        }
      }
    ]
  }
}
```

Kadrian:
So how would you propose 
1) the frontend should send an external query to the backend?
2) the backend resolve response should look like?

manuel-hegner:
The backend already expects:

```javascript
{
  type: "CONCEPT_QUERY",
  root: {
    type: "AND",
    children: [{
      type: "EXTERNAL"
      ...
    }]
  }
}
```

Leaving out the array is just a convention so that manually written json looks simpler (especially in other cases), but automatically created json should always use the additional brackets. We will also always print the brackets, even if not given in the input.

Kadrian:
We agreed to change what the frontend sends to

```
{
  type: "CONCEPT_QUERY",
  root: {
    type: "EXTERNAL",
    ...
  }
}
```

then the backend should respond to a GET `stored-queries/:id` with something that contains:
```
{
  type: "CONCEPT_QUERY",
  root: {
    type: "EXTERNAL_RESOLVED",
    ...
  }
}
```

PR #534, opened 2019-04-25
Feature/additional fields for selects and collecting of all namespaced ids

thoniTUB:
additional fields for selects - is required to determine a specific type for an external service (CATEGORICAL)

collecting of all namespaced ids - allows to collect all namesspaced id in a CQuery and hereby enable to easily switch the dataset


Kadrian:
Merged on behalf of @thoniTUB

PR #535, opened 2019-04-25
customizable  column names

thoniTUB:
makes csv column names customizable by using a provided function from a higher level.

This way the csv column names can be adjusted to the needs of the application, e.g long and verbose or short and distinct


backend/src/main/java/com/bakdata/conquery/models/query/concept/SelectResultInfo.java
```diff
@@ -0,0 +1,22 @@
...
+	private final Select select;
+
+	public SelectResultInfo(String name, ResultType type, Integer sameNameOcurrences, int postfix, Select select) {
+		super(name, type, sameNameOcurrences, postfix);
+		this.select = select;
+	}
+	
+	public SelectResultInfo withName(@NonNull String name) {
+	    if (name == null) throw new java.lang.NullPointerException("name");
+		return this.getName() == name ? this : new SelectResultInfo(name, this.getType(), this.getSameNameOcurrences(), this.getPostfix(), this.getSelect());
```
Kadrian: This is not nested inside the if, but looks like it, could you clean this up a little?

thoniTUB: Done. Check was already covered by annotation

Kadrian:
Merging on behalf of @thoniTUB

PR #543, opened 2019-04-29
changed versioning approach

manuel-hegner:


Kadrian:
Could you explain the change? I don't understand maven that well

Kadrian:
@manuel-hegner ?

PR #585, opened 2019-05-13
Feature/plugin configs

thoniTUB:
leaves a space in the config for pluggable external configurations, which implement the PluginConfig interface


thoniTUB:
I tried ClassToInstanceMap, but adding the classes to it in a loop didn't work as expected

PR #586, opened 2019-05-13
adds missing increment operation for columns that have the same name

thoniTUB:


backend/src/main/java/com/bakdata/conquery/apiv1/ResultCSVResource.java
```diff
@@ -51,7 +52,9 @@
 		.builder()
 		.prettyPrint(true)
 		.nameExtractor(
-			sd -> sd.getCqConcept().getIds().get(0).toStringWithoutDataset() + "_" + sd.getSelect().getId().toStringWithoutDataset())
+			sd -> ConqueryEscape.unescape(sd.getCqConcept().getIds().get(0).toStringWithoutDataset())
```
manuel-hegner: Unescaping here is a bad idea.  The toString() method is escaping the string for a reason.

PR #620, opened 2019-05-22
Try to make heroku deploy again.

Kadrian:
Hey @MarcusBaitz 

Could you quickly describe how the heroku deployment process works? 
#607 

Current status:
- There's a heroku secret configured in travis ci, that should allow pushing to heroku
- There's a constant error we're getting
```
$ bash frontend/scripts/deploy-heroku-docker.sh conquery-dev
Error response from daemon: login attempt to https://registry.heroku.com/v2/ failed with status: 401 Unauthorized
```
- The pipeline is NOT failing, even though there's an error.

(I've added `set -e` to abort script execution when there's an error on login)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ stages:
     if: branch = develop AND type = push
   - name: Heroku Master
     if: branch = master AND type = push
-  - name: "Prerelease"
+  - name: Prerelease
     if: type = push AND tag IS present AND tag =~ ^v\d+\.\d+(\.\d+)?-.*$
-  - name: "Release"
+  - name: Release
     if: type = push AND tag IS present AND tag =~ ^v\d+\.\d+(\.\d+)?$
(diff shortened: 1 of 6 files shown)


MarcusBaitz:
Hey @Kadrian,

the PR #607 fix: `WARNING! Using --password via the CLI is insecure. Use --password-stdin.`

In addition, we have a `401 Unauthorized`. For that, i renewed the Travis secret for Heroku. Locally it worked for me.

That the pipeline does not fail is correct. The artifact is still usable.

I would recommend to remove `set -e` to avoid side effects.

Kadrian:
What's the artifact used for, in it can't be pushed? Why is the pipeline still failing?

Kadrian:
@manuel-hegner Do you remember a reason why conquery in the production build would need to be built with the public path `/app/static`?

PR #624, opened 2019-05-23
Fix disappearing FE

thoniTUB:
the frontend does not handle null values, so an empty list is send instead


Kadrian:
Haha you're not fixing the "disappearing FE", you're fixing the API spec :-P

PR #674, opened 2019-06-18
Introduce up to date API types, refactor

Kadrian:


frontend/lib/js/api/types.js
```diff
@@ -0,0 +1,272 @@
+// @flow
+
+// This file specifies
```
Kadrian: You might want to check this file @manuel-hegner @thoniTUB @awildturtok

PR #768, opened 2019-07-17
Results visualization (first draft)

Kadrian:
That's a **first working draft**, which
- adds a button "Preview" next to the results view
- on clicking, downloads the CSV and shows a full-screen preview of the first 1000 lines.

### Todo

- [x] figure out whether 1000 makes sense
- [x] display a loading indication while CSV is downloading
- [x] visualize dates including gaps
- [ ] find a way to get the value types for each column from the backend cc @manuel-hegner 
- [ ] then calculate & show some preliminary statistics for each column


Kadrian:
Other todos in follow-up PRs

PR #832, opened 2019-08-26
Feature/make queries reexecutable

Kadrian:
This makes queries with a missing `resultUrl` re-executable.

The backend functionality is still missing, so this assumes that the following reexecute-endpoint works:

1) `POST /datasets/<datasetId>/queries/<storedQueryId>/reexecute`

( I would be open to calling this 
`POST /datasets/<datasetId>/stored-queries/<storedQueryId>/reexecute` )

The frontend now expects to find an `id` in the response payload. If that's found,

2) `GET /datasets/<datasetId>/stored-queries/<id>` is called, using that obtained `id`.

Step 2) is repeated (long polling) until `status` is either `FAILED` / `CANCELED` or `DONE`


thoniTUB:
@awildturtok you had a different approach on reexecution, i guess. One that does not require a new endpoint

awildturtok:
A query solely consisting of a REUSED_QUERY will automatically execute the prior query instead of creating a new query executing the old query. It is therefore not necessary to create or talk with a new endpoint, just sending a query is sufficient.

Kadrian:
@awildturtok I discussed this back then with Manuel. I guess what you mean by

> A query solely consisting of a REUSED_QUERY

is that you expect the frontend to make the following steps
1) Getting the stored query id from the endpoint that lists all stored queries 
2) making another GET for that stored query to get the "expanded" query, 
3) make another POST with that expanded query through the usual endpoint?

Manuel and I talked about that and agreed that it's not a lot of work for the backend to provide another re-execute endpoint which will allow us to save step 2) from above.

( @Priehn I'm not quite understanding why you approved the PR ;-) )

awildturtok:
@Kadrian no the Frontend would only have to send a query of the kind `{ type : SAVED_QUERY, query : $id}` to the query endpoint. The id should already be available in that context?

Kadrian:
@awildturtok 

So you're saying: To execute, do a
```
POST /datasets/<datasetId>/queries
```

using body
```
{ 
  "type" : "SAVED_QUERY", 
  "query" : "example.query.id.xyz"
}
```
?

That'd be fine with me as well, I'll just have adjust this PR a little bit then.
Is that functionality already live on the backend-side?

awildturtok:
```
{
    "type": "CONCEPT_QUERY",
    "root": {
        "query": "$query_id",
        "type": "SAVED_QUERY"
    }
}
```

thoniTUB:
Looks like this is almost done, right?

PR #865, opened 2019-10-08
Filter Search now be configured to have different search modes.

awildturtok:


backend/src/main/java/com/bakdata/conquery/apiv1/FilterSearch.java
```diff
@@ -12,16 +18,70 @@
...
+		/**
+		 * String must start with search-String.
+		 */
+		PREFIX {
+			@Override
+			double score(String candidate, String keyword) {
+				/* 0...1 depending on the length ratio */
+				double matchScore = (double) candidate.length() / (double) keyword.length();
+
+				/* boost by 1 if matches start of keyword */
```
thoniTUB: I think the boot comment is copy mistake

backend/src/main/java/com/bakdata/conquery/apiv1/FilterSearch.java
```diff
@@ -12,16 +18,70 @@
...
+	@Getter
+	public static enum FilterSearchType {
+		/**
+		 * String must start with search-String.
+		 */
+		PREFIX {
+			@Override
+			double score(String candidate, String keyword) {
+				/* 0...1 depending on the length ratio */
+				double matchScore = (double) candidate.length() / (double) keyword.length();
```
thoniTUB: So longer results are favored over short results, isn't it better to show the shortest matches first

PR #896, opened 2019-10-28
CQConcept skips unresolved connectors.

awildturtok:
This enables us to execute queries partially.


awildturtok:
great commit message!

PR #919, opened 2019-11-12
Upgrade a couple of dependencies

Kadrian:
This upgrades quite some dependencies.
Majorly:
- webpack
- babel

It removes `query-string` and replaces that with `URLSearchParams`, which is a nice standard web api that is supported by all browsers except for IE11. 

**To Support IE, the latest babel version should include a polyfill. But this needs to be tested**


Kadrian:
@awildturtok You agreed to test this with IE11! Nice! :-)

awildturtok:
Kloppt!

PR #965, opened 2019-12-15
Add forms endpoint to get configured forms

Kadrian:
This adds the endpoint to the other api endpoints, such that the user's bearer token will be used when forms are loaded. (Even though this endpoint won't be used from within conquery yet)

Exposes
- updateReducers (to replace the main reducer once the forms have been loaded)
- getForms - an endpoint to load the forms
- form config types to type the getForms response


thoniTUB:
@Kadrian ist der schon ready?

Kadrian:
Ja!

PR #978, opened 2020-01-06
Feature/fix remove user from groups on delete

thoniTUB:
when a user was deleted it reference was not removed from groups it was a member of.
this causes that the group object could not be loaded at restart.

this PR fixes that and improves the restart test


backend/src/main/java/com/bakdata/conquery/models/auth/entities/User.java
```diff
@@ -82,17 +82,18 @@ public UserId createId() {
...
+			if(roles.add(role)) {
+				log.trace("Added role {} to user {}", role.getId(), getId());
 				updateStorage(storage);
 			}
 		}
 	}
 	
+	@Override
 	public void removeRole(MasterMetaStorage storage, Role role) throws JSONException {
 		synchronized (roles) {
```
awildturtok: Synchronized weg überall bitte

PR #1001, opened 2020-01-23
check download permission on all dataset references in submitted query

thoniTUB:


backend/src/main/java/com/bakdata/conquery/models/query/Visitable.java
```diff
@@ -2,10 +2,8 @@
 
 import java.util.function.Consumer;
 
-import com.bakdata.conquery.models.query.concept.CQElement;
-
 public interface Visitable {
```
awildturtok:
```{java}
Visitable<T extends Visitable<?>> {

void visit(Consumer<T> visitor);

}

```

backend/src/main/java/com/bakdata/conquery/models/query/concept/HasNamespacedIds.java
```diff
@@ -0,0 +1,17 @@
...
+import com.bakdata.conquery.models.execution.ManagedExecution;
+import com.bakdata.conquery.models.identifiable.ids.NamespacedId;
+
+/**
+ * Classes that contain {@link NamespacedId}s on their first level should implement this interface.
+ * It can be used by the Visitor pattern that is implemented by {@link ManagedExecution} to collect all ids
+ * that were supplied by a User before the structure was initialized.
+ * These ids are than used to e.g. check permissions.
+ */
+public interface HasNamespacedIds {
```
awildturtok: NamespacedIdHolding/Having

PR #1004, opened 2020-01-27
Synchronize on bucket assignment and creation

awildturtok:
This will not immediately be merged as the other code creates better outputs but only blocks a bug, not completely avoiding the possible state.


backend/src/main/java/com/bakdata/conquery/models/jobs/ImportJob.java
```diff
@@ -95,13 +95,24 @@ public void execute() throws JSONException {
...
+					namespace.addResponsibility(bucket);
+				}
 			}
+
 			for (WorkerInformation w : namespace.getWorkers()) {
 				w.send(new UpdateWorkerBucket(w));
 			}
-			namespace.updateWorkerMap();
+
+			synchronized (namespace) {
```
thoniTUB: Hast du hiermit noch etwas vor?

PR #1019, opened 2020-02-06
SourceSearch is tracked in SlowJobQueue

awildturtok:


awildturtok:
Also no longer allocates an ImmutableMap in QuickSearch, as that was extremely slow (5min to each updateMatchingStats).

backend/src/main/java/com/bakdata/conquery/apiv1/FilterSearch.java
```diff
@@ -83,18 +83,13 @@ public double score(String candidate, String keyword) {
 
 	private static Map<String, QuickSearch<FilterSearchItem>> search = new HashMap<>();
 
-	public static ExecutorService init(Namespaces namespaces, Collection<Dataset> datasets) {
-		ExecutorService executor = Executors.newSingleThreadExecutor();
-
+	public static void init(Namespaces namespaces, Collection<Dataset> datasets, JobManager jobManager) {
```
awildturtok: doku

PR #1041, opened 2020-02-17
Realm checks if crediential list is not null

thoniTUB:


thoniTUB:
However config validation is not working as expected

backend/src/main/java/com/bakdata/conquery/apiv1/auth/PasswordCredential.java
```diff
@@ -17,5 +18,6 @@
 @AllArgsConstructor
 public class PasswordCredential implements CredentialType {
 
+	@NotEmpty
```
awildturtok: Doku bitte

backend/src/main/java/com/bakdata/conquery/models/auth/web/DefaultAuthFilter.java
```diff
@@ -70,8 +70,9 @@ public void filter(final ContainerRequestContext requestContext) throws IOExcept
 			// Success an extracted token could be authenticated
 			log.trace("Authentication was successfull for token type {}", token.getClass().getName());
 			return;
+
 		}
-		log.warn("Non of the configured realms was able to successfully authenticate the following token(s).");
+		log.warn("Non of the configured realms was able to successfully authenticate the extracted token(s).");
 		log.trace("The failing tokens were: {}", failedTokens);
```
awildturtok: kein token loggen bitte

PR #1046, opened 2020-02-19
Fix typo in TableView

Kadrian:


Kadrian:
@awildturtok Thanks for merging. When are the changes merged back into develop (+ who takes care?)

PR #1053, opened 2020-02-24
Check in permissions that the current dataset matches as well

Kadrian:
😅Forgot to check the current dataset match at the final permission check again. This fixes that!


Kadrian:
@thoniTUB Wann wird das denn in den develop zurückgemerged?

PR #1076, opened 2020-03-13
Make form concept node copyable

Kadrian:
Simple possibility to copy form concepts within a form concept to other fields, using drag & drop.

This is the first step. 

The next steps would be to make copying more fine-graular, as in:
1) make selected subsets of nodes copy-able
2) make all nodes copy-able


thoniTUB:
Sorry Kai, hab den falschen Branch geupdated.

PR #1080, opened 2020-03-16
Add a FormConceptCopyModal to select concepts to copy in detail

Kadrian:
Looks like this:
Bildschirmfoto 2020-03-16 um 15 41 58

And is almost done.

TODO:
- [ ] Make sure there is no conflict with expand-unexpand subnodes after a successful copying
- [ ] Refactor this a bit to make the logic less complicated (e.g. no nested reduce)


Kadrian:
@thoniTUB Was ist hier passiert?

PR #1089, opened 2020-03-18
Expand form subnodes right after their parents

Kadrian:
Instead of just appending the expanded nodes at the end


thoniTUB:
Changed base branch to integrate fix into next release

PR #1120, opened 2020-03-24
Typescript

Kadrian:
I've realised that before I can start building bigger features, typescript would make that process way safer and faster.

So this moves the frontend from `flow` types to `typescript` and `create-react-app`.

## 🎉 Why - the advantages
You could say this upgrades the project from 2017 to 2020.

Typescript brings **heavy** advantages:

- Way **better maintainability**, especially as the codebase grows and grows nowadays. Less worries about breaking code when adding new features.
- Great tooling helps developing faster and safer, **reducing the small bugs** we have had to fix later.
- Way easier and **faster onboarding** experience for anyone who would join this project at a later point in time.

Create-react-app also brings great advantages:
- We don't have to maintain our own webpack config anymore, including quirks and tweaks to make this work with localization. You can say we just moved one abstraction layer higher.
- We got rid of ~20 lower-level direct dependencies.
- Since create-react-app is actively maintained, when there are any performance upgrades to the build process in the future, we get them for free.

## 🔥 Breaking changes
@thoniTUB @awildturtok  

- The compiled files now reside in `frontend/build`, not in `frontend/dist` anymore.
- Right now the build contains no separate `index.de.html` and `index.en.html` anymore, but only a single `index.html`, depending on the `REACT_APP_LANG` env value it was built with.
- The environment parameters have been renamed (due to `create-react-app` standards), the following are supported now:

```
    NODE_ENV: "development" | "production";
    REACT_APP_API_URL: string;
    REACT_APP_DISABLE_LOGIN: boolean;
    REACT_APP_LANG: "de" | "en";
    PORT: number;
```

## 📖 How to move forward
- I've set up `yarn typecheck` and `yarn lint` commands now which we can use in the CI, to finally get feedback on the frontend code as well.
- Right now we've got 2460 typecheck errors, which we'll have to resolve step by step. They're not there because the code isn't working, but because it's not sufficiently well typed yet. I've set `TSC_COMPILE_ON_ERROR=true` so the code still compiles to Javascript, and everything works like expected. The plan is now to reduce these errors down to 0 and then to remove this flag. But the good thing is: This can be done bit by bit now.

## 📌 TODO before this can be merged
- [x] Fix the `jest` test setup
- [x] Fix the maven build setup according to the breaking changes


thoniTUB:
Spricht etwas dagegen, dass sich das Frontend auf die Locale des User anpasst?

frontend-resources/pom.xml
```diff
@@ -23,7 +23,7 @@
 	<build>
 		<resources>
 			<resource>
-				<directory>${workingDirectory}/app/dist/</directory>
+				<directory>${workingDirectory}/build/</directory>
```
thoniTUB: In #1120  you mentioned  a change of the compile output directory. The change here does not exactly mirror this. Is that correct anyways?

Kadrian: Yes, It was actually even under `/app/...`, didn't mention that above

Kadrian:
> Spricht etwas dagegen, dass sich das Frontend auf die Locale des User anpasst?

Not really. But then we'd also need to allow switching the language on the fly when you realize you wanted the other one. So  => future work

PR #1160, opened 2020-04-15
fix concurrent modification exception on permission cleanup

thoniTUB:


backend/src/main/java/com/bakdata/conquery/tasks/QueryCleanupTask.java
```diff
@@ -142,52 +142,46 @@ public static int deleteQueryPermissionsWithMissingRef(MasterMetaStorage storage
...
-						if (wpermission.getInstances().size() == validRef.size()) {
-							// All are valid, nothing changed proceed with the next permission
-							continue;
-						}
-						// Create a new Permission that only contains valid references
-						WildcardPermission reducedPermission = new WildcardPermission(
-							List.of(wpermission.getDomains(), wpermission.getAbilities(), validRef), wpermission.getCreationTime());
-						owner.addPermission(storage, reducedPermission);
+				}
+				if (validRef.size() > 0) {
```
awildturtok, line 170:
```suggestion
				if (!validRef.isEmpty()) {
```

PR #1189, opened 2020-05-12
Feature/form config fix

thoniTUB:
- Fixes a bug where the label is overriden on creation of the formConfig
- Allowes to patch values member of form config
- saves the creation time of the form config


backend/src/main/java/com/bakdata/conquery/models/auth/permissions/Ability.java
```diff
@@ -8,15 +8,18 @@
...
 	SHARE,
 
-	// Dataset Specific
+	// Dataset specific
 	DOWNLOAD,
-	PRESERVE_ID
+	PRESERVE_ID,
+	
+	// FormConfig specific
+	MODIFY
```
Kadrian, line 22: 🤔 😬

Kadrian: Just curious: Why is a new permission necessary? If it really is, how about calling it MODIFY_FORM_CONFIG and removing the comment?

backend/src/main/java/com/bakdata/conquery/util/VariableDefaultValue.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+
+/**
+ * Marker annotation for the AutoDoc project to signal that this fields default value is not stable because it is e.g. randomly generated or the current time.
```
awildturtok, line 11: Sollte das dann nicht ins Autodoc Projekt? kP wie die dependencies sind

thoniTUB: Ich gebe dir recht. Das sollten wir in einem Projekt-Refactor machen, in dem z.B. diese Annotation in einem extra Projekt ist, das von beiden (AutoDoc und Backend) importiert wird

backend/src/main/java/com/bakdata/conquery/models/auth/permissions/Ability.java
thoniTUB: The new Permission is the FormConfigPermission. This is just one of the abilities, that is available for that type.
A stricter differentiation between Meta-Data of an object (creation.-time, tags, ...) and the actual content might be a better approach to also simplify the Permission and their abilities.

Kadrian: Okay, so: Why is the new "FormConfigPermission" necessary?

And about the naming of the abilities here, it seems a little odd that we have TAG LABEL SHARE for queries, but for formConfigs it's called `MODIFY`, no?

thoniTUB: We have a Permission for each type of protected Resource (Query, Dataset, Formular ...).
The abilities you mentioned also apply to the FormConfigPermission, those were initially only for the QueryPermission. However the possiblity to modify the content of an Query does not exsists at the moment.
The Abilities are probably a bit to fine grain at this point.

PR #1219, opened 2020-06-05
Shortens title for export formular

thoniTUB:


backend/src/main/java/com/bakdata/conquery/models/config/PreprocessingConfig.java
```diff
@@ -3,15 +3,13 @@
...
-import org.hibernate.validator.constraints.NotEmpty;
-
 import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
 
 @Getter @Setter @ToString
 public class PreprocessingConfig {
-	@NotEmpty @Valid
+	@Valid
```
Kadrian, line 12: Hmm is that change intended?

thoniTUB: Yes, since the preprocess task has been updated to optionally accept these informations as a comand line argument. The Config is then ignored

PR #1236, opened 2020-06-24
ExistsAggregator only return True when filters are true

awildturtok:
Old ExistsAggregator can also be true when aggregating filters are in the chain and false.


backend/src/main/java/com/bakdata/conquery/models/query/concept/specific/CQConcept.java
```diff
@@ -149,17 +154,24 @@ private long calculateBitMask(ConceptElement<?>[] concepts) {
...
+		List<ExistsAggregator> existsAggregators = new ArrayList<>();
 
 		for (Select s : select) {
 			AggregatorNode<?> agg = new AggregatorNode<>(s.createAggregator());
+
+			if(agg.getAggregator() instanceof ExistsAggregator){
+				existsAggregators.add(((ExistsAggregator) agg.getAggregator()));
+			}
+
 			plan.addAggregator(agg.getAggregator());
```
thoniTUB, line 175:
```suggestion
		List<AggregatorNode<?>> nodes = new ArrayList<>();

		for (Select s : select) {
			AggregatorNode<?> agg = new AggregatorNode<>(s.createAggregator());

			plan.addAggregator(agg.getAggregator());
```

PR #1244, opened 2020-06-26
adds pageination

thoniTUB:
`curl 'http://localhost:8080/api/datasets/adb_bosch/stored-queries?limit=2&offset=1'`
Returns the Status of the second and third newest Query.


thoniTUB:
Alright, was a quick idea, but i know that it conflicts with the search

PR #1265, opened 2020-07-10
Start mock backend with start script.

### LINKED ISSUE #1264: Demo backend has to be started by hand
The [docs](https://github.com/bakdata/conquery/blob/d1193499ab04b89d9c544b9db93b4bd018d5778a/README.md#starting-the-demo) on how to start the demo are incomplete. The backend has to be started with `yarn serve` (which btw is the wrong command as it makes the user think that this would somehow serve the frontend code, not a mock-backend). Easiest fix is to just start the backend with the `yarn start` command. I see no point in starting a dev preview for this app without a running backend....

Levino:
Make start script also start the mock backend for easier dev setup. Fixes #1264.


Levino:
I made this a draft in order to prevent merging. We need to update the docs since #1272 has been merged into `develop`.

Issue #1262, opened 2020-07-10
Typescript code does not build without errors


Levino:
When I try to run `yarn start` I get a lot of typescript errors in the console. It looks like the webserver with the react app is running anyhow, which is a bit surprising.

Levino:
So running `yarn typecheck` results in 1831 errors.

The command `yarn typecheck` should be run in CI and block PRs from being merged if failling.

awildturtok:
Hi Levino, thanks for your suggestions!

The change to Typescript was (Afaik) meant to be a step by step migration instead of a complete makeover. The frontend is as of yet not integrated into our backend pipeline.

Levino:
A transition is in order. Ignoring typescript errors however is not. You want to go easy on the screws and tighten them one after the other. But once any screw is fixed, any error caused by it must block code from being merged.

For example in this case here one could set `implicitAny` to `true` and turn a lot of the errors into warnings. But a non-existent import would still be an error (which must be an error because if I import something which does not exist that is worse than not importing anything! And these errors currently exist, see #1267).

So I suggest to relax the rules for typescript and fix the errors that are still reported until only warnings are reported and then step by step tighten the rules and fix the then warnings-turned-errors. Btw: Fixing most of the errors (90%) should be doable in an hour or so because it is mostly about missing types.

Levino:
And one question remains which could potentially be answered by @Kadrian: Why is it possible to build? Does the typescript step run with `transpileOnly`? Why do these errors not block the build?

awildturtok:
https://github.com/bakdata/conquery/blob/develop/frontend-resources/pom.xml#L63

This POM documents the steps we take to build and package our frontend maybe that might help?

For testing purposes I usually run `yarn` then `yarn start` which is sufficient.

Levino:
So this is what I get for the build:
```bash
$ yarn build
yarn run v1.22.4                                                                                                                                                                                                  
$ react-app-rewired build                                                                                                                                                                                         
Creating an optimized production build...
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
[...]

PR #1267, opened 2020-07-10
Fix wrongly named import.

Levino:
Fixing one of the 1832 errors referred to in #1262.

I do not understand how this app can even build. Are the errors from unused code?


awildturtok:
Thank you for taking the time to fix this!

PR #1272, opened 2020-07-15
Feature/fix docs and types for fe

awildturtok:


awildturtok:
> Also these changes (once fixed) should be backported to the branch deliverable for obvious reasons...

I personally agree with you, but we've decided against this as it could lead to people having different states of the repo. In hindsight, we should have made it a tag or communicated a commit hash as that was our intention to begin with.

PR #1359, opened 2020-09-22
use concept tree child labels in column name

thoniTUB:
This adds more information to the default columnName if no label was set by the user.

Previously only the concept name was set, even if a concept child was chosen. Now both are displayed


awildturtok (commented):
Kannst du dafür einen Test schreiben?

awildturtok (approved):

PR #1360, opened 2020-09-22
Fix queries on concepts without validity date

thoniTUB:
Previously those queries failed due to a NullPtrExp


awildturtok (approved):
nice!

PR #1417, opened 2020-11-03
use request provided locale to format numbers in result csv

thoniTUB:


Kadrian:
Does this change the API?

thoniTUB:
@Kadrian Numbers of the resulting csv is formated according to the `Accept-Language`-Header of the download request.
Does the frontend preview use a fixed locale for number parsing?
If numbers contain digit grouping is independent from this PR. It influences however, how they are presented when they are activated (QueryParam `pretty=true` [default])

Kadrian:
Currently, the frontend preview expects money to be decimal separator `,` and maybe thousand separator `.`, but once we can use `pretty=false`, I'd change it to simply parse it as as a float of "cents". 

Does that make sense? I think it's a good idea to use the accept-language header for the pretty=true / default case.

awildturtok (approved):

awildturtok (approved):

PR #1430, opened 2020-11-10
call a private method instead of creating a lambda

thoniTUB:
Reduce creation of lambda objects

```
 num     #instances         #bytes  class name (module)
-------------------------------------------------------
   1:     608625161    56831278992  [Ljava.lang.Object; (java.base@11.0.8)
....
com.bakdata.conquery.models.common.daterange.CDateRangeExactly
  19:      19069613      457670712  com.bakdata.conquery.models.types.parser.specific.string.StringParser$$Lambda$402/0x00007efe435ff0b0
  20:       7113683      227637856  it.unimi.dsi.fastutil.bytes.ByteArrayList
```

### DIFF
--- a/backend/src/main/java/com/bakdata/conquery/models/types/parser/specific/string/StringParser.java
+++ b/backend/src/main/java/com/bakdata/conquery/models/types/parser/specific/string/StringParser.java
@@ -62,7 +62,7 @@ public class StringParser extends Parser<Integer> {
 
 	@Override
 	protected Integer parseValue(String value) throws ParsingException {
-		return strings.computeIfAbsent(value, this::parseNewValue);
+		return strings.computeIfAbsent(value, this::processNewValue);
 	}
 	
 	@Override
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/bakdata/conquery/models/types/parser/specific/string/StringParser.java
```diff
@@ -49,19 +49,20 @@
 	public StringParser(ParserConfig config) {
 
 	}
+	
+	private int parseNewValue(String value) {
```
awildturtok, line 53:
```suggestion
	private int processNewValue(String value) {
```

backend/src/main/java/com/bakdata/conquery/models/types/parser/specific/string/StringParser.java
```diff
@@ -49,19 +49,20 @@
...
+		//return next id
+		return strings.size();
+	}
 
-			//return next id
-			return strings.size();
-		});
+	@Override
+	protected Integer parseValue(String value) throws ParsingException {
+		return strings.computeIfAbsent(value, this::parseNewValue);
```
awildturtok, line 65:
```suggestion
		return strings.computeIfAbsent(value, this::processNewValue);
```

awildturtok (approved):

PR #1452, opened 2020-11-27
allow relative formqueries to only set features or outcomes

thoniTUB:
Sometimes one is not interested in both. But one has to be set!


backend/src/main/java/com/bakdata/conquery/apiv1/forms/DateContextMode.java
```diff
@@ -27,11 +28,12 @@
 	COMPLETE(null){
 		@Override
 		public List<CDateRange> subdivideRange(CDateRange range) {
-			return List.of(range);
+			return range == null ? ImmutableList.of() : List.of(range);
```
awildturtok, line 31:
```suggestion
			return range == null ? Collections.emptyList() : List.of(range);
```

awildturtok (dismissed):

PR #1455, opened 2020-11-30
Ignore .env by default, create a .env.example; add details to README

Kadrian:


frontend-resources/pom.xml
```diff
@@ -80,6 +80,16 @@
 							<arguments>install --frozen-lockfile --non-interactive --ignore-engines</arguments>
 						</configuration>
 					</execution>
+					<execution>
+						<id>copy .env</id>
+						<goals>
+							<goal>cp</goal>
```
Kadrian, line 86: Well, somehow this won't work :-( Quite annoying that copying a single file with maven is so tricky.

Means that this PR is probably on hold, until we manage to make frontend pipelines work without maven.

Kadrian:
Closing this in favor of https://github.com/bakdata/conquery/pull/1619 which also needs more ENV variables and makes similar changes

PR #1472, opened 2020-12-03
Prevent awaitDone to run through unfinished execution

thoniTUB:
Previously some test where successful but caused exceptions because their assertion was met and the cleanup began before the execution of the query finished. The collect-results-job then failed to find the user.


thoniTUB:
@awildturtok sorry habe mich verklickt und schon gemerged

PR #1510, opened 2020-12-16
don't send PrimaryDict to shards

awildturtok:


backend/src/main/java/com/bakdata/conquery/models/jobs/ImportJob.java
```diff
@@ -280,10 +280,6 @@ private boolean createMappings(PreprocessedHeader header) throws JSONException {
 
 			getProgressReporter().report(1);
 
-			log.debug("\t\tsending");
-
-			namespace.sendToAll(new UpdateDictionary(primaryDict));
-
 			getProgressReporter().report(1);
```
thoniTUB, line 283: NOOP progress :D

thoniTUB (approved):

PR #1534, opened 2021-01-08
Bump java version to 14

awildturtok:


awildturtok:
This does not work with Azure at the moment, maybe switch to Github runners before the switch

PR #1596, opened 2021-02-01
Feature/create exists in export form

awildturtok:


backend/src/main/java/com/bakdata/conquery/apiv1/forms/export_form/ExportForm.java
```diff
@@ -86,6 +89,16 @@ public void collectNamespacedIds(Set<NamespacedId> ids) {
 	public void resolve(QueryResolveContext context) {
 		timeMode.resolve(context);
 		prerequisite = Form.resolvePrerequisite(context, queryGroup);
+
+		// createExists is always true in ExportForm.
+		this.visit(visitable -> {
+			if(visitable instanceof CQAnd){
+				((CQAnd) visitable).setCreateExists(true);
```
Kadrian, line 96: Curious: So you can't use a `||` in the if, because of casting?

Kadrian:
Looks good to me – what about other forms?

backend/src/main/java/com/bakdata/conquery/apiv1/forms/export_form/ExportForm.java
awildturtok: It's not a shared subclass/interface and static typing doesn't allow what you propose (structural like TS would probably allow it).

awildturtok:
@Kadrian the other forms are either obsolete through the jupyter forms, or Jupyter Forms which just issues Export forms.

backend/src/main/java/com/bakdata/conquery/apiv1/forms/export_form/ExportForm.java
thoniTUB: But a shared Interface might be the right call here

thoniTUB (dismissed):

PR #1613, opened 2021-02-09
properly handle potential empty Column in DateRangeParser

awildturtok:


thoniTUB (approved):

backend/src/main/java/com/bakdata/conquery/models/events/parser/specific/DateRangeParser.java
```diff
@@ -82,8 +82,15 @@ protected DateRangeStore decideType() {
 		}
 
 		// They need to be aligned.
```
awildturtok, line 84:
```suggestion
		// They need to be aligned if they are non-empty.
```

PR #1615, opened 2021-02-09
add CompressionFilter to Mina IPC

awildturtok:
This might make result collection faster, I'll test that first


awildturtok:
This slows down query transmission significantly

PR #1619, opened 2021-02-11
Feature/add keycloak support

Kadrian:
Introduces keycloak support by setting the environment variable `REACT_APP_IDP_ENABLE=true`

**NOTE:**  From now on, setting up the frontend requires cloning the `.env.example` file:

```bash
cp .env.example .env
```

Then adjust the variables in `.env` if necessary.

To build and start the frontend via docker, you can then:

```bash
./scripts/start_production.sh
```


Kadrian:
Frontend tests failing is to be expected, we'll need to adjust the CI to clone the `.env.example` file

PR #1667, opened 2021-03-02
Chunk reader using async parser

awildturtok:


awildturtok:
Does not change anything drastic on huge (4million) datasets. Still 3s. Maybe improvements between machines, but not necessaruky,

awildturtok:
Does not improve load times

PR #1674, opened 2021-03-05
Align form config sharing with previous query sharing

Kadrian:


Kadrian:
cc @thoniTUB this would still require some backend api changes before it can work

PR #1715, opened 2021-03-18
Refactor translations, use i18next

Kadrian:
i18next is way more modern, compared to the old i18n-react that we've
been using. i18next also comes with a hooks api and typescript support,
reducing our typescript errors by ~120

I really got warm with this one :-D


thoniTUB:
I'm looking into why the CI broke

thoniTUB:
Does that mean that language isn't a compile time thing any more?

thoniTUB:
@Kadrian ready to merge?

Kadrian:
Nice, thanks for fixing the CI.

> Does that mean that language isn't a compile time thing any more?

Well if you mean that we don't have to use the LANG env variable anymore, that's true in a way:

- Switching languages using a button in the app could now happen during runtime
- Adding text from different sources (e.g. a translator software) during compile time or runtime is easier now
- It makes translations more statically analyzable, so finding missing translations e.g. could become a compile time thing soon.
- i18next automatically warns when encountering missing translations during runtime.

PR #1900, opened 2021-06-01
Use streaming xlsx format

thoniTUB:
Uses the apache POI streaming format SXSSF (instead of XSSF):

# Benchmark
## Table 
 - ~260000 rows 
 - ~31 column
- CSV-size: ~800 MB (takes some time to download as well)

## Download
- Rendering + Transfer

|         | Time | Size |
| --- | --- | --- |
| prev | ~14 min | 230 MB| 
| now | ~2 min | 240 MB| 

# Misc
- Auto Column size does not work atm, which is leading to pound-fills: 
image
- Still the rendering is performed first and afterwards the transfer starts
- Medium table download time is on par with csv due to smaller transfer size (csv: 150 MB; xlsx:38 MB )
- Small tables download "instantly"


awildturtok (approved):
Wo landen die tmp daten? Kannst du das bitte noch checken?
Aber nice

PR #1979, opened 2021-07-08
fix arrow file size explosion

thoniTUB:


thoniTUB:
Need to revise this as i cannot read the produced files using the python library

awildturtok (approved):

PR #2198, opened 2021-11-09
FIX checks if any aggregators have been set

thoniTUB:


backend/src/main/java/com/bakdata/conquery/models/query/queryplan/ConceptQueryPlan.java
```diff
@@ -155,6 +155,9 @@ public boolean isOfInterest(Entity entity) {
 	}
 
 	public boolean isAggregateValidityDates() {
+		if (aggregators.isEmpty()) {
```
Njimefo, line 158: gut

Njimefo (approved):

awildturtok (approved):

PR #2308, opened 2022-01-12
feature/shorter column names if possible

thoniTUB:


thoniTUB:
Closed for because the requirement changed

PR #2340, opened 2022-01-28
Allow If-None-Match header for CORS requests, to cache GET requests

Kadrian:


Kadrian:
cc @thoniTUB I noticed this one was also still missing

PR #2350, opened 2022-02-02
adds basic cypress test

thoniTUB:


frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,82 @@
+/// <reference types="cypress" />
+
+context('Visit conquery as users with different permissions', () => {
```
Kadrian, line 3: We're using prettier to auto-format js / ts code on save. Could you configure your editor to use prettier? (and use our installed / configured prettier – see `package.json`)

frontend/cypress.json
```diff
@@ -0,0 +1 @@
+{}
```
Kadrian, line 1: Is this file required?

frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,82 @@
...
+            ).as('meInfo')
+    
+            cy.visit("http://localhost:3000/?access_token=user.user2")
+    
+            cy.wait(['@config', '@datasets', '@meInfo'])
+        })
+
+        
+        it('Can see dataset', () => {
+            cy.get('#downshift-0-input').invoke('attr','value').should('eq',"Dataset1")
```
Kadrian, line 75: Time for me to start adding proper test ids in the future

scripts/run_backend_cypress.sh
```diff
@@ -0,0 +1,24 @@
...
+config=$(cat << EOF
+ {
+    "server": {
+        "applicationConnectors": [
+            {
+                "type": "http",
+                "port": "8080"
+            }
+        ],
+		"adminConnectors": [
```
Kadrian, line 13: How about cleaning the indentation a bit here? :-D

Kadrian (commented):
Cool draft so far! Would it make sense to add some instruction to README.md as well?

frontend/cypress/integration/user_frontend_test.js
thoniTUB: I added that to the ReadMe aswell

frontend/cypress.json
thoniTUB: Not jet, but thats where the `baseUrl` is supposed to be i guess ;)

frontend/README.md
```diff
@@ -80,6 +80,15 @@ When queried for login:
...
 
+**Code Formating**  [![code style: prettier](https://img.shields.io/badge/code_style-prettier-ff69b4.svg?style=flat-square)](https://github.com/prettier/prettier)
+
+The frontend JavaScript/TypeScript code is formatted using `prettier`.
+It can be invoked on the command line through `yarn`:
+```
+yarn prettier --write /path/to/file
+```
+or convieniently through a plugin in VS Code (see https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode).
+
```
Kadrian, line 91:
````suggestion
**Code formatting**  [![code style: prettier](https://img.shields.io/badge/code_style-prettier-ff69b4.svg?style=flat-square)](https://github.com/prettier/prettier)

The frontend TypeScript code is formatted using `prettier`.

We recommend you configure your editor to auto-format on save. If you're using VS-Code, for example, there's a plugin: [Prettier – Code formatter](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode).

You could also invoke prettier on the command line:
```
yarn prettier --write /path/to/file
```

````

File: frontend/src/js/pane/Pane.tsx [hunk 6]
```diff
@@ -24,16 +24,23 @@ interface PropsT {
...
     <Root className={className}>
       <Container>
-        <PaneTabNavigation tabs={tabs} paneType={paneType} />
-        <Container>{children}</Container>
+        <PaneTabNavigation
+          tabs={tabs}
+          paneType={paneType}
+          dataTestId={dataTestId}
+        />
+        <Container data-test-id={dataTestId + "Container"}>
```
Kadrian, line 41:
```suggestion
        <Container data-test-id={dataTestId + "-container"}>
```

frontend/src/js/query-runner/QueryRunnerButton.tsx
```diff
@@ -66,7 +66,7 @@ const QueryRunnerButton = ({
   const icon = getIcon(isStartStopLoading, isQueryRunning);
 
   return (
-    <StyledBasicButton type="button" onClick={onClick} disabled={disabled}>
+    <StyledBasicButton type="button" onClick={onClick} disabled={disabled} data-test-id="queryRunnerButton">
```
Kadrian, line 69:
```suggestion
    <StyledBasicButton type="button" onClick={onClick} disabled={disabled} data-test-id="query-runner-button">
```

frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,95 @@
...
+        url: "/api/me",
+      }).as("meInfo");
+
+      cy.visit("http://localhost:3000/?access_token=user.user1");
+
+      cy.wait(["@config", "@datasets", "@meInfo"]);
+    });
+
+    it("Cannot see concepts", () => {
+      cy.get('[data-test-id="leftPane"]').contains("Konzepte");
```
Kadrian, line 25:
```suggestion
      cy.get('[data-test-id="left-pane"]').contains("Konzepte");
```

frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,95 @@
...
+
+      cy.contains("Keine Konzepte");
+    });
+
+    it("Cannot see queries", () => {
+      cy.get('[data-test-id="leftPane"]').contains("Anfragen").click();
```
Kadrian, line 31:
```suggestion
      cy.get('[data-test-id="left-pane"]').contains("Anfragen").click();
```

frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,95 @@
...
+
+      cy.contains("Keine Anfragen gefunden");
+    });
+
+    it("Cannot see forms", () => {
+      cy.get('[data-test-id="leftPane"]').contains("Formulare").click();
```
Kadrian, line 37:
```suggestion
      cy.get('[data-test-id="left-pane"]').contains("Formulare").click();
```

frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,95 @@
...
+        url: "/api/me",
+      }).as("meInfo");
+
+      cy.visit("http://localhost:3000/?access_token=user.user2");
+
+      cy.wait(["@config", "@datasets", "@meInfo"]);
+    });
+
+    it("Can see dataset", () => {
+      cy.get('[data-test-id="datasetSelector"]')
```
Kadrian, line 64:
```suggestion
      cy.get('[data-test-id="dataset-selector"]')
```

frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,95 @@
...
+        .should("eq", "Dataset1");
+    });
+
+    it("Can see concepts", () => {
+      cy.contains("Concept1");
+    });
+
+    it("Can execute query", () => {
+
+      cy.get('[data-test-id="rightPaneContainer"]')
```
Kadrian, line 76:
```suggestion
      cy.get('[data-test-id="right-pane-container"]')
```

frontend/cypress/integration/user_frontend_test.js
```diff
@@ -0,0 +1,95 @@
...
+
+      cy.contains("Concept1").trigger("dragstart").trigger("dragleave");
+      cy.get('@queryEditor')
+        .trigger("dragenter")
+        .trigger("dragover")
+        .trigger("drop")
+        .trigger("dragend");
+
+      cy.get('@queryEditor')
+        .find('[data-test-id="queryRunnerButton"]').click()
```
Kadrian, line 89:
```suggestion
        .find('[data-test-id="query-runner-button"]').click()
```

Kadrian:
Here's a PR that merges in develop and resolves the conflicts in package.json / yarn.lock: https://github.com/bakdata/conquery/pull/2367

frontend/package.json
```diff
@@ -115,10 +116,6 @@
     "ts-node": "^10.5.0",
     "typescript": "^4.5.5"
   },
-  "resolutions": {
```
Kadrian, line 118: Did you remove this intentionally?

thoniTUB: Yes, I wanted to ask you about that, since it caused a problem in the install stage, when i invoked yarn with `--frozen-lockfile`:
```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```
Does these packages need to have fixed versions?

Kadrian: That's an unrelated error. Yes those versions should be fixed. I'd suggest you
- add them back
- run `yarn` without `--frozen-lockfile`
- commit the changes to yarn.lock

thoniTUB: The thing is, that yarn.lock won't change if i run it without `--frozen-lockfile`. I tried it, and run `yarn --ignore-platform --frozen-lockfile` afterwards again and the Problem remains. Or do you suggest, that `--frozen-lockfile` should never be used?

thoniTUB: I got the idea from this: https://github.com/yarnpkg/yarn/issues/4778#issuecomment-339448000

Kadrian: We're only using `--frozen-lockfile` in CI, so we can make sure that the dependencies installed in an automated CI build are exactly the ones we expect

thoniTUB: That was my intention. You can try to run `yarn --ignore-platform --frozen-lockfile` on the `develop` branch. That's where i am getting the error too. For some reason the `Test Frontend`-Workflow does not use the frozen-flag. Why is that so?

thoniTUB: Out of curiosity: what makes these Package version so important to stay at that fixed version?

thoniTUB: @Kadrian I kept the resolutions and adjusted the version of babel-loader in the dependencies to be the same as in the resolutions. This works for me now.

backend/src/main/java/com/bakdata/conquery/models/config/ColumnConfig.java
```diff
@@ -77,40 +77,34 @@
 	 *
 	 * Also Name of output column for {@link FrontendConfig.UploadConfig#getIdResultInfos()}, ergo output csv-Columns.
 	 */
-	@InternalOnly
 	private String field;
```
thoniTUB, line 80: @awildturtok Weißt du warum hier `@InternalOnly` dranne war? Das Problem war hier das die Werte beim kopieren der Config in Standalone nicht mitgenommen wurden

frontend/cypress/support/backend_config.json
```diff
@@ -0,0 +1 @@
+{}
```
thoniTUB, line 1: this is empty for now, but in case that a configuration is needed the place is here

Kadrian (changes requested):
Looking pretty good! But I tried to run it locally and it doesn't work yet – I left you some comments

scripts/run_frontend_cypress.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/env bash
+proj_dir=$(readlink -f `dirname $0`/..)
+
+pushd "$proj_dir/frontend"
+
+npx serve -s build -l 3000
```
Kadrian, line 6: What does this do, by the way? Can / should we build & start the frontend using the Dockerfile, in a container?

README.md
```diff
@@ -18,10 +18,36 @@ Conquery supplies a powerful interface to group event types in a hierarchical *c
...
+- Maven 3 (optional for building)
+- Java JDK 11
+- Node.js 16 + Yarn
+
+#### **Steps**
+To test frontend and backend together you can start the setup that is used for end-to-end tests.
+
+First build the backend using `scripts/build_version.sh` or download a JAR from the [release page](https://github.com/bakdata/conquery/releases).
+
+Then start frontend and backend by running `scripts/run_backend_cypress.sh`.
```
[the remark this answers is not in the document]
thoniTUB: I'll revise that, thank you!
There is a script call `run_conquery_cypress.sh` that runs both

scripts/run_frontend_cypress.sh
thoniTUB: This spins up an adhoc HTTP file-server serving files from the folder `build` (similar to pythons `python3 -m http.server` ). I will add that to the documentation, that the frontend need to be build beforehand. `npx` download `serve` incase it was not installed. I wouldn't add `serve` as a dev dependency, since no actual dev dependencies are needed here. 

Using Docker poses an extra requirement towards the need to having Docker installed and thus having to invoke a `sudo`-action at least once. All other tools like maven, java, node usually don't need `sudo`.

So if you just want to test the app or run the cypress test, you couldn't do so in a restricted environment.

scripts/load_test_data_cypress.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/env bash
+proj_dir=$(readlink -f `dirname $0`/..)
```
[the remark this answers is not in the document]
thoniTUB: I guess `-f` can be removed. It just follows symlinks

scripts/run_frontend_cypress.sh
```diff
@@ -0,0 +1,8 @@
...
+
+popd
```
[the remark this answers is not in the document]
thoniTUB: In my opinion, the scripts behavior is more consistent this way. I would not expect that my current directory changes after i run a random script/command

scripts/run_frontend_cypress.sh
Kadrian: > Using Docker poses an extra requirement towards the need to having Docker installed and thus having to invoke a sudo-action at least once. All other tools like maven, java, node usually don't need sudo.

Alright, I mean we should try to replicate a real world as close as possible but I'd also prefer a simple setup.

But then I'd vote to not add another dependency but just use `server.js`, which was written precicely to serve the built frontend from `./build`

frontend/src/js/app/RightPane.tsx
```diff
@@ -12,10 +12,11 @@ interface PropsT {
   tabs: TabT[];
 }
 
+// Shrink to size 0 so jQuery/cypress can differentiate between :hidden and :visible
 const Tab = styled("div")<{ isActive: boolean }>`
-  height: 100%;
+  height: ${(isActive) => (isActive ? "100%" : 0 )};
```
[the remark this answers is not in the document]
thoniTUB: Cypress uses jQuery under the hood, but i could find a way to filter DOM elements by their style/css directly. The most convienient way seem to use jQueries `:visible/:hidden` selector, which only looks at widths and heights of the elements:
https://api.jquery.com/visible-selector/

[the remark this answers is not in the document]
thoniTUB: Do you mean the container that holds the Dropzone and the QueryEditor? We can do that, but from a user handling perspective the visibility "flag" or some thing like it is better. 

I haven't tested it, but i imagine that if i reference the pane by it's id rather than its visibility, it is possible that can d'n'd elements in the QueryEditor eventhough its tab isn't actually in the foreground. This way an error might slip through.

What do you think of that?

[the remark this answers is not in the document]
thoniTUB: I think i tried that without success, but basically that is what is done in this video: https://www.youtube.com/watch?v=LxkrhUEE2Qk So I'll try that again, i must have missed something the first time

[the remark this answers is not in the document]
thoniTUB: I was able to fix that :) 
The selector is got simpler now

scripts/load_test_data_cypress.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/env bash
```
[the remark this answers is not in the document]
thoniTUB: @Kadrian würde bei dir auch `#!/bin/bash` gehen?

scripts/load_test_data_cypress.sh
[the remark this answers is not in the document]
thoniTUB: Okay, `-f` is needed, since is outputs the actual path. 

Unfortunately I haven't found a good portable way. There is however a greadlink in homebrew coreutils. Is that an option?

scripts/load_test_data_cypress.sh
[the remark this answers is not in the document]
Kadrian: Yep!

scripts/load_test_data_cypress.sh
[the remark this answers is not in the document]
Kadrian: I've added a suggestion that works for me in here https://github.com/bakdata/conquery/pull/2400 (simple avoids using readlink)

.github/workflows/test-cypress.yml
```diff
@@ -0,0 +1,41 @@
...
+          TSC_COMPILE_ON_ERROR: true
+          REACT_APP_API_URL: "http://localhost:8080"
+          REACT_APP_DISABLE_LOGIN: true
+          REACT_APP_LANG: de
+          REACT_APP_IDP_ENABLE: false
+          REACT_APP_IDP_REALM: Myrealm
+          REACT_APP_IDP_CLIENT_ID: frontend
+        with:
+          working-directory: ./frontend
+          build: yarn run react-app-rewired build
```
thoniTUB, line 38: Anstatt des commands hier ein script benutzen, damit das bauen lokal gleich funktioniert

Kadrian:
Todo: 
- remove readlink

thoniTUB:
add separate .env.e2e that is sourced for the cypress frontend

Kadrian:
Another todo:
- [ ] avoid adjusting different kinds of pom.xml versions when building for cypress

<details>
<summary>Screenshots showing the issue</summary>
Bildschirmfoto 2022-03-22 um 14 21 58
Bildschirmfoto 2022-03-22 um 14 21 49
</details>

Kadrian:
are there some chmod +x missing?

Bildschirmfoto 2022-03-28 um 10 51 27

README.md
```diff
@@ -16,12 +14,54 @@ Event data sets typically associate events with a certain subject (i.e. a person
...
 This repository includes the Conquery frontend along with a non-functional backend. It provides a set of example concept trees to demonstrate the capabilities of the UI: The example's use case is to search for groups of actors who appeared in movies of the same genre or received the same award.
 
 Check the README in `/frontend` for details.
 
+### Frontend + Backend
+
+#### Steps
+To test frontend and backend together you can start the setup that is used for end-to-end tests.
+
+First build the backend using `conquery/scripts/build_version.sh` or download a JAR from the [release page](https://github.com/bakdata/conquery/releases) and place it in `conquery/executable/target/`.
```
Kadrian, line 36:
```suggestion
First build the backend using `./scripts/build_version.sh` or download a JAR from the [release page](https://github.com/bakdata/conquery/releases) and place it in `./executable/target/`.
```

README.md
```diff
@@ -16,12 +14,54 @@ Event data sets typically associate events with a certain subject (i.e. a person
...
+
+Build the frontend by running:
+```
+$ cd frontend
+$ cp .env.example .env
+$ yarn build
+```
+
+You can then run `conquery/scripts/run_conquery_cypress.sh` to start frontend and backend, and also load the test data required by cypress end-to-end test or you can run `conquery/scripts/run_backend_cypress.sh` and `conquery/scripts/run_frontend_cypress.sh` separately without loading any data.
```
Kadrian, line 45:
```suggestion
You can then run `./scripts/run_conquery_cypress.sh` to start frontend and backend, and also load the test data required by cypress end-to-end test or you can run `./scripts/run_backend_cypress.sh` and `./scripts/run_frontend_cypress.sh` separately without loading any data.
```

README.md
```diff
@@ -16,12 +14,54 @@ Event data sets typically associate events with a certain subject (i.e. a person
...
+The frontend is accessible at http://localhost:8000 as the default "superuser" implicitly. Since the backend uses a development authentication, you can switch users by passing another users "UserId" as the access token in the query string when accessing the frontend, e.g.: http://localhost:8000/?access_token=user.user2.
+
+## Development
+
+### Testing
+Apart from separate frontend and backend tests, the project also contains end-to-end tests powered by cypress. 
+
+To run the end-to-end test locally:
+1. Make sure you installed all [requirements](#requirements)
+2. From the repo root folder run  `conquery/scripts/run_conquery_cypress.sh`
```
Kadrian, line 58:
```suggestion
2. From the repo root folder run `./scripts/run_conquery_cypress.sh`
```

README.md
```diff
@@ -16,12 +14,54 @@ Event data sets typically associate events with a certain subject (i.e. a person
...
+3. Wait until the output: `Node server listening on port: 8000` appears
+4. Start cypress from the `conquery/frontend` folder in another terminal: `$ yarn cypress open`
+5. Then chose a test suite and start it.
```
Kadrian, line 61: This 5-steps-description is great! Tested & works like this for me :)

frontend/.env.e2e
```diff
@@ -0,0 +1,23 @@
+# THIS IS THE ENV-FILE USED FOR CYPRESS TESTS
```
Kadrian, line 1: 👍

Kadrian (commented):
Thanks for the remaining changes! I think this PR is good to go, once you've `chmod +x`-ed the new scripts. 

I've added relative path suggestions for the README, please feel free to take or ignore those

Kadrian (dismissed):

README.md
thoniTUB: I thought about that, but I decided to be explicit because there are multiple `scripts` folder

thoniTUB:
@Kadrian i had to tinker a bit. Unfortuanatelly i could not use all the scripts in the ci, just the start script

PR #2400, opened 2022-03-09
Suggest a version of the run scripts that work for me

Kadrian:


Kadrian:
@thoniTUB See [9c5b39d](https://github.com/bakdata/conquery/pull/2400/commits/9c5b39dc9fb536e005fdea0da94ffad818c59f4c) for the scripts change, the rest is just merging in develop again, fixing yarn.lock conflicts.

Feel free to merge!

scripts/run_backend_cypress.sh
```diff
@@ -1,7 +1,6 @@
-#!/bin/env bash
-script_dir=`dirname $0`
+#!/bin/bash
 
-jar="$(readlink -f $script_dir/../executable/target/executable*.jar)"
-config="$(readlink -f $script_dir/../frontend/cypress/support/backend_config.json)"
+jar="../executable/target/executable*.jar"
+config="../frontend/cypress/support/backend_config.json"
 
 java -jar $jar standalone $config
```
Kadrian, line 6: Okay, it still won't work for me :-/ 

```
The operation couldn’t be completed. Unable to locate a Java Runtime.
Please visit http://www.java.com for information on installing Java.
```

But I ran `brew install openjdk` and now this is what it looks like on my side:

Bildschirmfoto 2022-03-09 um 17 31 56

@thoniTUB Which makes me wonder: how about actually using docker to build & run the local backend / frontend?

thoniTUB: We cannot run docker on our Virtual Desktops 😢

thoniTUB: I first thought the problem is, that java is just an alias. Since they are not available by default inside a script.

thoniTUB: Can you check that you installed openjdk-jdk instead of openjdk-jre?

thoniTUB: And confirm that `JAVA_HOME` is set correctly

PR #2410, opened 2022-03-22
Improve share modal behavior

Kadrian:
- Close after successful share (was more of a bug)
- Indicate loading state
- Disable button as long as no user groups have changed.

### DIFF
--- a/frontend/src/js/previous-queries/list/ShareProjectItemModal.tsx
+++ b/frontend/src/js/previous-queries/list/ShareProjectItemModal.tsx
@@ -1,12 +1,13 @@
 import styled from "@emotion/styled";
-import { StateT } from "app-types";
-import { useEffect, useState } from "react";
+import type { StateT } from "app-types";
+import { useEffect, useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useSelector } from "react-redux";
 
 import type { SelectOptionT, UserGroupT } from "../../api/types";
 import PrimaryButton from "../../button/PrimaryButton";
 import { TransparentButton } from "../../button/TransparentButton";
+import FaIcon from "../../icon/FaIcon";
 import Modal from "../../modal/Modal";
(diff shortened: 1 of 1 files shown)


frontend/src/js/previous-queries/list/ShareProjectItemModal.tsx
```diff
@@ -110,28 +117,32 @@ const ShareProjectItemModal = ({ item, onClose }: PropsT) => {
       ? t("sharePreviousQueryModal.unshare")
       : t("common.share");
 
+  const buttonDisabled =
+    JSON.stringify(initialUserGroupsValue) === JSON.stringify(userGroupsValue);
```
awildturtok, line 121: This seems rather hacky to me, relying on the serialization of the state object? Wouldn't it be preferable to use an elementwise comparison instead (likely also much more efficient)

awildturtok (commented):
It seems array equality is a [contentious topic](https://stackoverflow.com/questions/7837456/how-to-compare-arrays-in-javascript) in the JS community. I'd personally still rely on a less magic way, or at least encode it explicitly in a helper method (which can easily be plugged to use "better" variants should they be released). But it's not a mjaor problem.

frontend/src/js/previous-queries/list/ShareProjectItemModal.tsx
Kadrian: I understand your point, but user groups are short arrays, this is efficient enough, so I'd rather avoid micro optimizations here

awildturtok: My point isn't as much about optimization than about misappropriating serialization for something as basic as equality checks. But if you'd rather keep it I'm fine with it.

Kadrian: Agreed that it feels weird until you've seen it lots of times in JS code. It's more common than you'd think: https://stackoverflow.com/a/1144249/921486

awildturtok (approved):

PR #2413, opened 2022-03-22
Fix config folder filter to work exactly like query filter

Kadrian:


frontend/src/js/external-forms/form-configs/selectors.ts
```diff
@@ -63,15 +63,20 @@ export const selectFormConfigs = (
   folders: string[],
   noFoldersActive: boolean,
 ) => {
-  if ((!searchTerm || searchTerm.length === 0) && filter === "all") {
+  const noFilterSet =
+    (!searchTerm || searchTerm.length === 0) &&
+    filter === "all" &&
+    folders.length === 0 &&
+    !noFoldersActive;
```
Kadrian, line 70: This is the relevant change, the rest is formatting

awildturtok: Minor point: Wouldn't it make more sense to extract a common function for noFiltersSet at this point?

awildturtok (approved):
I'd personally the noFiltersSet logic into a helper method to avoid having to track both places, but it's not a dealbreaker: They are likely in their final form this way?

Kadrian:
Yeah, it's a good point, and there's more logic that can be abstracted for both FormConfig + PreviousQuery, but I'd like to wait at least until the new feature settles a bit.

PR #2422, opened 2022-03-29
Fix a bunch of TS errors, now about 30 remaining

Kadrian:


frontend/src/js/common/helpers/fileHelper.ts
```diff
@@ -1,8 +1,11 @@
 export const readFileAsText = (file: File) =>
-  new Promise((resolve, reject) => {
+  new Promise<string>((resolve, reject) => {
     const reader = new FileReader();
 
-    reader.onload = (evt) => resolve(evt.target.result);
+    reader.onload = (evt) =>
+      evt.target
```
awildturtok, line 6: Sometimes you use exists and sometimes not, is there an explicit pattern behind that?

frontend/src/js/external-forms/transformQueryToApi.ts
```diff
@@ -1,21 +1,23 @@
...
   return elementGroups.map(({ concepts, connector, ...rest }) =>
     concepts.length > 1
       ? {
           type: connector,
-          children: transformElementsToApi(concepts),
+          children: transformElementsToApi(concepts.filter(exists)),
           ...rest,
         }
-      : { ...transformElementsToApi(concepts)[0], ...rest },
+      : { ...transformElementsToApi(concepts.filter(exists))[0], ...rest },
```
awildturtok, line 20: Couldn't this cause an Index out of Bounds? Or is it ensured that you'll have existing concepts

frontend/src/js/preview/actions.ts
```diff
@@ -24,9 +24,9 @@ export function useOpenPreview() {
     dispatch(loadCSVForPreview.request());
 
     try {
-      const parsed = await loadCSV(url);
+      const result = await loadCSV(url);
 
-      dispatch(loadCSVForPreview.success({ csv: parsed.result.data, columns }));
+      dispatch(loadCSVForPreview.success({ csv: result.data, columns }));
```
awildturtok, line 29: Not something for this PR, but I'd advise you to look into stream processing (using the step parameter) of the csv (For example if we are going to do a redesign or add features).

It could probably be much more snappy/responsive if you didn't have to store the entire CSV in memory for preview, especially since as far as I know only the first 500 or so lines are actually preivewable. It can sometimes freeze the whole frontend for bigger files

awildturtok (approved):
The only questionable thing is filtering concepts, but I'd guess it's not an actual problem.

frontend/src/js/common/helpers/fileHelper.ts
Kadrian: Yes, `exists` is mostly used as a type guard. In this case it's not needed, ternary auto-casts into boolean

frontend/src/js/preview/actions.ts
Kadrian: Check 👍

awildturtok (dismissed):

frontend/src/js/external-forms/transformQueryToApi.ts
Kadrian: Done!

Kadrian: Could have actually led to a bug if you tried to only fill the second or-condition of an otherwise optional concept field. So good catch!

PR #2432, opened 2022-04-05
Refactoor: remove dynamic reducers per tab

Kadrian:
Brings us down from 32 TS errors to 29 😭✌️

There should be no change in features

But at least removes A LOT of older code, removes the (very complex ) abstraction that tabs could have dynamically updated reducers, because this has become unnecessary.


frontend/src/app-types.ts
```diff
@@ -1,10 +1,3 @@
...
-import type { ExternalFormsStateT } from "./js/external-forms";
-import type { StandardQueryEditorStateT } from "./js/standard-query-editor";
-import type { TimebasedQueryEditorStateT } from "./js/timebased-query-editor";
 
-export interface StateT extends BaseStateT {
-  queryEditor: StandardQueryEditorStateT;
-  timebasedQueryEditor: TimebasedQueryEditorStateT;
-  externalForms: ExternalFormsStateT;
-}
+export interface StateT extends BaseStateT {}
```
Kadrian, line 3: I'll add a follow-up refactoring soon to remove this file

awildturtok (approved):

PR #2438, opened 2022-04-07
make ids URL-save

thoniTUB:


thoniTUB:
Paused/discarded because Browsers automatically decode Codes that do not need to be encoded, which conflicts with our id splitting

PR #2440, opened 2022-04-11
Use Autosaving forms when starting form queries

Kadrian:
In turn, remove any manual FormConfig saving


frontend/src/js/external-forms/FormConfigLoader.tsx
```diff
@@ -0,0 +1,181 @@
...
+      if (typeof value === "string") {
+        return datasetOptions.find((option) => option.value === value);
+      }
+
+      return value;
+    case "SELECT":
+      if (typeof value === "object") return value as SelectOptionT;
+      if (typeof value === "string") {
+        const options = field.options.map((option) => ({
+          label: option.label[activeLang] || "",
```
awildturtok, line 72: Wouldn't it make more sense to use any fallback value (eg. the first value in label: `values(option.labels).shift()`) instead of displaying nothing?

awildturtok: Caveat being that I don't know if you generally fallback onto value and if it'd be preferable to accidentally display spanish instead of a potentially weird technical term.

awildturtok (approved):

PR #2479, opened 2022-05-10
Feature/improve e2e tests

Kadrian:


scripts/run_e2e_backend.sh
```diff
@@ -3,10 +3,17 @@
 jar="./executable/target/executable-0.0.0-SNAPSHOT.jar"
 config="./frontend/cypress/support/backend_config.json"
```
thoniTUB, line 4: @Kadrian fails currently the test
```suggestion
config="./cypress/support/backend_config.json"
```

cypress/integration/test_1_runQuery.js
```diff
@@ -0,0 +1,36 @@
...
+
+const USER_TOKEN_WITH_PERMISSIONS = "user.user2";
+
+describe("Run query", () => {
+  before(() => {
+    visitWithToken(USER_TOKEN_WITH_PERMISSIONS);
+  });
+
+  it("Can execute query and see it in the queries tab", () => {
+    cy.get('[data-test-id="right-pane-container"] >div:visible').as(
```
awildturtok, line 12: Syntactic question, why does Cypress do naming here instead of just using a variable that you can reference?

scripts/run_e2e_frontend.sh
```diff
@@ -2,16 +2,24 @@
 set -e
 
 
+echo ""
```
awildturtok, line 5: Echo without params should always give you a newline.
```suggestion
echo
```

awildturtok (approved):

scripts/run_e2e_frontend.sh
Kadrian: done for all of those 👍

cypress/integration/test_1_runQuery.js
Kadrian: Ah, because it's good practice to always **select** elements right before doing something with them

PR #2492, opened 2022-05-16
Send creatable=true if no searches are available

awildturtok:
Merge after: https://github.com/bakdata/conquery/pull/2491

In favor of: https://github.com/bakdata/conquery/pull/2485

### DIFF
--- a/backend/src/main/java/com/bakdata/conquery/apiv1/frontend/FEFilter.java
+++ b/backend/src/main/java/com/bakdata/conquery/apiv1/frontend/FEFilter.java
@@ -37,11 +37,11 @@ public class FEFilter {
 	private List<FEValue> options = Collections.emptyList();
 
 	/**
-	 * min value for range selects.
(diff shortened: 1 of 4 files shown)


thoniTUB (dismissed):
Danke für die Doku

backend/src/main/java/com/bakdata/conquery/apiv1/frontend/FEFilter.java
```diff
@@ -19,19 +19,53 @@
...
 	@Builder.Default
 	private List<FEValue> options = Collections.emptyList();
+
+	/**
+	 * min value for range selects.
+	 */
 	private Integer min;
+	/**
+	 * max value for range selects.
+	 */
```
awildturtok, line 45:
```suggestion

	/**
	 * min value for range filters.
	 */
	private Integer min;
	/**
	 * max value for range filters.
	 */
```

PR #2506, opened 2022-05-23
Allow tooltips for concept filters

Kadrian:


frontend/mock-api/concepts.json
```diff
@@ -138,6 +141,7 @@
             {
               "id": "budget",
               "label": "Budget",
+              "tooltip": "A money range tooltip, let's see how that works.",
```
awildturtok, line 144:
```suggestion
              "tooltip": "A money range tooltip, let's see how that works.\nMaybe it could benefit from newlines?",
```

awildturtok (changes requested):
Is tooltip a field you hook into in general? so is simply filling/exposing it sufficient for it to be displayed?

Minor nit: In general we use `feature/*` style branches, and try to reserve top-level names for release/master/develop etc. Please try to also conform to that style unless you see important reasons not to.

frontend/mock-api/concepts.json
awildturtok: If not possible applicable, the PR can be merged

Kadrian: Hehe, that won't work unfortunately

Kadrian:
> Is tooltip a field you hook into in general? so is simply filling/exposing it sufficient for it to be displayed?

Yes, as soon as a filter supplies `tooltip`, this will work. It's an optional field though, can be left out from object (= undefined)

> Minor nit: In general we use feature/* style branches, and try to reserve top-level names for release/master/develop etc. Please try to also conform to that style unless you see important reasons not to in the future.

Wow that you noticed! I did that on purpose, because I'm trying to loosen up that "feature/" restriction, but let's discuss this in person

frontend/mock-api/concepts.json
Kadrian: What will work though is `Check this<br />newline`

Kadrian:
@awildturtok any other requested changes to the PR?

frontend/mock-api/concepts.json
awildturtok: Changed it.

awildturtok (dismissed):
Nothing in terms of functionality, I'd like to see the break inside one tooltip go into the tests, but user your own judgment!

PR #2535, opened 2022-06-07
Add initial history UI

Kadrian:


frontend/src/js/entity-history/actions.ts
```diff
@@ -65,68 +76,131 @@ export const initHistoryData = createAsyncAction(
...
+      return first && last
+        ? {
+            ...row,
+            dates: {
+              from: first,
+              to: last,
+            },
+          }
+        : row;
+    })
```
awildturtok, line 192: There are valid rows, with only either min or max.

awildturtok (approved):

PR #2566, opened 2022-06-28
reverts runner image

thoniTUB:


thoniTUB:
@SebChmie reverting the image for now until we can test the alpine image thoroughly

PR #2590, opened 2022-07-11
Improve history UI

Kadrian:


frontend/src/js/entity-history/Timeline.tsx
```diff
@@ -104,126 +134,258 @@ const Box = styled("div")`
...
+    const columnBuckets = useMemo(() => {
+      return {
+        money: columns.filter((c) => c.type === "MONEY"),
+        concepts: columns.filter(
+          (c) =>
+            c.semantics.length > 0 && c.semantics[0].type === "CONCEPT_COLUMN",
+        ),
+        secondaryIds: columns.filter(
+          (c) =>
+            c.semantics.length > 0 && c.semantics[0].type === "SECONDARY_ID",
```
awildturtok, line 181: I wouldn't bet on there being only 1 semantic per column.

PR #2591, opened 2022-07-11
Fix expand query big multi select filter values

Kadrian:


frontend/src/js/api/types.ts
```diff
@@ -69,16 +69,18 @@ export interface MultiSelectFilterT extends MultiSelectFilterBaseT {
...
   type: "BIG_MULTI_SELECT";
   // Not needed in this format:
-  template: {
-    filePath: string; // "/.../import/stable/Referenzen/example.csv",
-    columns: string[];
-    columnValue: string; // Unclear, what that's even needed for
-    value: string;
-    optionValue: string;
-  };
+  // template: {
```
Kadrian, line 77: @awildturtok is this still relevant? Does the API still send this?

PR #2598, opened 2022-07-11
Only use EN in history

Kadrian:


frontend/src/js/entity-history/actions.ts
```diff
@@ -150,7 +150,7 @@ export function useUpdateHistorySession() {
       }
 
       const authorizedCSVUrl = getAuthorizedUrl(csvUrl);
-      const csv = await loadCSV(authorizedCSVUrl);
+      const csv = await loadCSV(authorizedCSVUrl, { english: true });
```
awildturtok, line 153: Since semantics should tell you the things you need to know to make programatic decisions, you could even just use the user-agents lang. This might be helpful when passing the csv to the users with Datumswerte/dates and sources.

PR #2648, opened 2022-08-04
add isBusy endpoint to test if any known work is being done

awildturtok:


thoniTUB (approved):
https://github.com/bakdata/conquery/pull/2643

PR #2665, opened 2022-08-15
History iteration 3 b

Kadrian:


frontend/src/js/entity-history/Timeline.tsx
```diff
@@ -419,14 +275,113 @@ export const Timeline = memo(
...
+      const lastEvt = groupedEvents[groupedEvents.length - 1][0];
+
+      const isDuplicateEvent = JSON.stringify(evt) === JSON.stringify(lastEvt);
+
+      if (isDuplicateEvent) {
+        continue;
+      }
+
+      const datesMatch =
+        JSON.stringify(evt.dates) === JSON.stringify(lastEvt.dates);
```
awildturtok, line 327: isn't dates always `[$start,$end]` at this point? You could probably shave some time by directly comparing the values.

frontend/src/js/entity-history/Timeline.tsx
```diff
@@ -419,14 +275,113 @@ export const Timeline = memo(
...
+      const sourcesMatch = evt.source === lastEvt.source;
+      const allSecondaryIdsMatch = secondaryIds.every(
+        ({ label }) => evt[label] === lastEvt[label],
+      );
+
+      const similarEvents = datesMatch && sourcesMatch && allSecondaryIdsMatch;
+
+      if (similarEvents) {
+        groupedEvents[groupedEvents.length - 1].push(evt);
```
awildturtok, line 336: Since grouping is done by equality of a set of keys, you could use something like the following:

```ts
let groups  = new Map<[],EntityEvent[]>

for (event in events) {
  let key = [evt.source, event.dates].concat(secondaryIds.map(id => evt[id])]
  if (!(key in groups)) {
    groups[key] = []
  }
  groups[key].push(evt)
}
```

Kadrian: Hmm I'm sorry I don't understand what you're saying

> Since grouping is done by equality of a set of keys
grouping is done by equality of a set of values, not keys here

what problem are you trying to address?

frontend/src/js/entity-history/Timeline.tsx
Kadrian: good point, let me change that

PR #2734, opened 2022-09-29
split preview connectors into available and default selection

awildturtok:


backend/src/main/java/com/bakdata/conquery/models/datasets/PreviewConfig.java
```diff
@@ -85,6 +88,14 @@ public InfoCardSelect(String label, SelectId select) {
 
 	}
 
+
+	@JsonIgnore
+	@ValidationMethod(message = "Default Connectors must also be available Connectors.")
+	public boolean isDefaultSubsetOfAvailable() {
+		return Sets.difference(getAllConnectors(), getDefaultConnectors()).isEmpty();
```
awildturtok, line 95:
```suggestion
		return Sets.difference(getDefaultConnectors(), getAllConnectors()).isEmpty();
```

PR #2795, opened 2022-11-08
Support big multi selects with initial options

awildturtok:


Kadrian:
👍  sorry for using the wrong base

Kadrian (approved):

PR #2805, opened 2022-11-15
Better reference list drop ux

Kadrian:


frontend/src/js/upload-concept-list-modal/UploadConceptListModal.tsx
```diff
@@ -131,24 +376,84 @@ const UploadConceptListModal = ({ onAccept, onClose }: PropsT) => {
...
-    useSelectConceptRootNodeAndResolveCodes();
-
-  if (!conceptCodesFromFile || conceptCodesFromFile.length === 0) {
+  if (!fileRows || fileRows.length === 0) {
     onClose();
   }
 
-  const hasUnresolvedItems = unresolvedItemsCount > 0;
-  const hasResolvedItems = resolvedItemsCount > 0;
+  const onSubmit = useCallback(
```
awildturtok, line 383: If you're not happy with this implementation, we could try moving the distinction of concepts/filters into the backend for the purpose of resolving?

awildturtok (approved):

frontend/src/js/upload-concept-list-modal/UploadConceptListModal.tsx
Kadrian: Hmm I know this looks a bit clunky, but I think it's actually okay, I will have to distinguish in the frontend in either case.

PR #2821, opened 2022-11-23
Automatically scrolls down to dropzone when element is dropped

fabian-bizfactory:


Kadrian (changes requested):
We have forms with multiple DropzoneLists in a single form, so I think using a single ref and exporting it might be a problem. Could you try with a `forwardRef` on DropzoneList?

Looks good otherwise!

frontend/src/js/external-forms/form-components/DropzoneList.tsx
```diff
@@ -1,5 +1,5 @@
 import styled from "@emotion/styled";
-import { ReactNode } from "react";
+import React, { ReactNode } from "react";
```
Kadrian, line 2: nit: if possible, let's try and avoid React default imports

fabian-bizfactory:
I tried it with forward ref
Snippet of DropZoneList
```TS
const DropzoneList = forwardRef(
  <DroppableObject extends PossibleDroppableObject>(
    props: WithFileProps<DroppableObject> | WithoutFileProps<DroppableObject>, ref:any) =>
  {
```
However, I get the error:
```
ERROR in src/js/external-forms/form-concept-group/FormConceptGroup.tsx:194:9
TS2322: Type '(item: DragItemFile | DragItemConceptTreeNode) => void' is not assignable to type '((props: PossibleDroppableObject, monitor: DropTargetMonitor<unknown, unknown>) => void) | ((props: PossibleDroppableObject, monitor: DropTargetMonitor<...>) => void)'.
  Type '(item: DragItemFile | DragItemConceptTreeNode) => void' is not assignable to type '(props: PossibleDroppableObject, monitor: DropTargetMonitor<unknown, unknown>) => void'.
    Types of parameters 'item' and 'props' are incompatible.
      Type 'PossibleDroppableObject' is not assignable to type 'DragItemFile | DragItemConceptTreeNode'.
        Type 'DragItemQuery' is not assignable to type 'DragItemFile | DragItemConceptTreeNode'.
          Type 'DragItemQuery' is missing the following properties from type 'DragItemConceptTreeNode': ids, tables, selects, tree, and 2 more.
```
Do you have any idea how to fix that issue? ( I did not want to commit something with errors, lmk if I should do it anyways)

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -112,6 +112,17 @@ const FormConceptGroup = (props: Props) => {
   const newValue = props.newValue;
   const defaults = props.defaults || {};
 
+  // indicator if it should be scrolled down back to the dropZone
+  const [scrollToDropzone, setScrollToDropzone] = useState<boolean>(false);
+  const dropzoneRef = createRef<HTMLDivElement>();
```
Kadrian, line 117: `createRef` creates a ref every time it's called, so in this case, it will create a new ref on every rerender of FormConceptGroup. Is there a reason you'd go for `createRef` instead of `useRef`? Otherwise, let's use `useRef` please

fabian-bizfactory: No particular reason, changed it to `useRef`

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -154,6 +165,7 @@ const FormConceptGroup = (props: Props) => {
   return (
     <div>
       <DropzoneList /* TODO: ADD GENERIC TYPE <ConceptQueryNodeType> */
+        ref={dropzoneRef == undefined ? null : dropzoneRef}
```
Kadrian, line 168: Why wouldn't `ref={dropzoneRef}` work?

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -112,6 +112,17 @@ const FormConceptGroup = (props: Props) => {
...
 
+  // indicator if it should be scrolled down back to the dropZone
+  const [scrollToDropzone, setScrollToDropzone] = useState<boolean>(false);
+  const dropzoneRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    if (scrollToDropzone) {
+      dropzoneRef.current?.scrollIntoView({ behavior: "smooth" });
+      setScrollToDropzone(false);
+    }
+  });
```
Kadrian, line 123: Why a useEffect without dependency array? If I understand correctly that will run the effect on every render. Then we don't even need the useEffect, right?

fabian-bizfactory: Yes, that is correct, fixed that

PR #2827, opened 2022-11-29
Activate Tabs by hovering an element over them

fabian-bizfactory:
This is implemented for:
- The big Tab Navigation
- Small Tabs


frontend/src/js/small-tab-navigation/SmallTabNavigation.tsx
```diff
@@ -26,7 +26,7 @@ const SmallTabNavigation: FC<PropsT> = ({
   onSelectTab,
 }) => {
   return (
-    <div className={className}>
+    <div className={className} style={{"display": "flex"}}>
```
Kadrian, line 29: What's the reason for that change? The parent components are currently setting layout settings like this. Plus, if we style, let's please try and avoid inline styles and use styled components instead

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,41 @@
...
+// 600 ms is the default time for a tooltip to appear
+// feels very responsive
+// can be overriden by passing a timeUntilNavigateOverride prop
+const TIME_UNTIL_NAVIGATE = 600;
+
+export const HoverNagiatable:FC<PropsT> = ({ timeUntilNavigateOverride, triggerNavigate, children, disabled }) => {
+    const time_to_navigate = timeUntilNavigateOverride || TIME_UNTIL_NAVIGATE;
+    let timeout: number | null | NodeJS.Timeout; // timeout can either be a number or a NodeJS.Timeout depending on the environment
+    const [_, drop] = useDrop({
+        accept: [DNDType.FORM_CONFIG, DNDType.CONCEPT_TREE_NODE, DNDType.PREVIOUS_QUERY, DNDType.PREVIOUS_SECONDARY_ID_QUERY],
```
Kadrian, line 22: Configuring prettier will also format those lines

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+// default time until the hover triggers navigation
+// 600 ms is the default time for a tooltip to appear
+// feels very responsive
+// can be overriden by passing a timeUntilNavigateOverride prop
+const TIME_UNTIL_NAVIGATE = 600;
+
+export const HoverNagiatable:FC<PropsT> = ({ timeUntilNavigateOverride, triggerNavigate, children, disabled }) => {
+    const time_to_navigate = timeUntilNavigateOverride || TIME_UNTIL_NAVIGATE;
+    let timeout: number | null | NodeJS.Timeout; // timeout can either be a number or a NodeJS.Timeout depending on the environment
```
Kadrian, line 20: why did you add that comment? Depending on which environment? browser vs server? We're only running the app in the browser

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,41 @@
+import { FC } from "react";
+import { useDrop } from "react-dnd";
+import { DNDType } from "../common/constants/dndTypes";
+
+
+interface PropsT {
+    timeUntilNavigateOverride?: number;
```
Kadrian, line 7: Let's not add props that we don't use. Even though it's a nice thought for the future, if we're never passing `timeUntilNavigateOverride` anywhere, let's remove the prop

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,41 @@
...
+const TIME_UNTIL_NAVIGATE = 600;
+
+export const HoverNagiatable:FC<PropsT> = ({ timeUntilNavigateOverride, triggerNavigate, children, disabled }) => {
+    const time_to_navigate = timeUntilNavigateOverride || TIME_UNTIL_NAVIGATE;
+    let timeout: number | null | NodeJS.Timeout; // timeout can either be a number or a NodeJS.Timeout depending on the environment
+    const [_, drop] = useDrop({
+        accept: [DNDType.FORM_CONFIG, DNDType.CONCEPT_TREE_NODE, DNDType.PREVIOUS_QUERY, DNDType.PREVIOUS_SECONDARY_ID_QUERY],
+        hover: (_, monitor) => {
+            // uses property of null => if null == false
+            if (timeout == null && !disabled) {
```
Kadrian, line 25: What's the reason for a non-explicit equality check? Let's please use explicit equality checks, so `===` instead. (we should configure this as an eslint rule)

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,41 @@
...
+// can be overriden by passing a timeUntilNavigateOverride prop
+const TIME_UNTIL_NAVIGATE = 600;
+
+export const HoverNagiatable:FC<PropsT> = ({ timeUntilNavigateOverride, triggerNavigate, children, disabled }) => {
+    const time_to_navigate = timeUntilNavigateOverride || TIME_UNTIL_NAVIGATE;
+    let timeout: number | null | NodeJS.Timeout; // timeout can either be a number or a NodeJS.Timeout depending on the environment
+    const [_, drop] = useDrop({
+        accept: [DNDType.FORM_CONFIG, DNDType.CONCEPT_TREE_NODE, DNDType.PREVIOUS_QUERY, DNDType.PREVIOUS_SECONDARY_ID_QUERY],
+        hover: (_, monitor) => {
+            // uses property of null => if null == false
```
Kadrian, line 24: What do you mean by that comment? Could you clarify this?

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,41 @@
+import { FC } from "react";
+import { useDrop } from "react-dnd";
+import { DNDType } from "../common/constants/dndTypes";
+
+
+interface PropsT {
+    timeUntilNavigateOverride?: number;
+    triggerNavigate: () => void;
+    disabled?: boolean;
```
Kadrian, line 9: Same for `disabled`

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
Kadrian, line 20: What happens when the component rerenders? Will the timeout just get overridden? I think we should probably use a timeout ref instead of a local variable

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// default time until the hover triggers navigation
+// 600 ms is the default time for a tooltip to appear
+// feels very responsive
+// can be overriden by passing a timeUntilNavigateOverride prop
+const TIME_UNTIL_NAVIGATE = 600;
+
+export const HoverNagiatable:FC<PropsT> = ({ timeUntilNavigateOverride, triggerNavigate, children, disabled }) => {
```
Kadrian, line 18: Going forward, let's try and avoid React.FC, see https://github.com/facebook/create-react-app/pull/8177. Preferrably, let's just inline props definitions like `const Comp = ({ a, b }: { a: string, b: number}) => {`

Kadrian (changes requested):
Alright, so far so good. I think the approach to use a `useDrop` hover + `monitor.isOver()` is good. I added some comments.

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
fabian-bizfactory: There are two reasons, why I added the NodeJS.Timeout (sorry for the misleading comment)
- Typescript does not like it if the Type is missing

frontend/src/js/small-tab-navigation/SmallTabNavigation.tsx
fabian-bizfactory: image
Breaks due to the Div in HoverNavigatable. But I will remove the style here

frontend/src/js/standard-query-editor/QueryNodeActions.tsx
```diff
@@ -18,7 +18,7 @@ const StyledFaIcon = styled(FaIcon)`
 `;
 
 const StyledIconButton = styled(IconButton)`
-  padding: 0px 6px 4px;
+  padding: 3.5px 6px 0px;
```
Kadrian, line 21: Hmm, how is this change related? Let's please avoid half-pixel values

Kadrian (changes requested):
When I tested this
- I can't see a hover indication when hovering over a tab, shouldn't there be one?
- the tab switch stops working after the switch happened once. So I can't hover Editor -> Form Editor -> Back to Editor. Could you look into that?

frontend/src/js/standard-query-editor/QueryNodeActions.tsx
fabian-bizfactory: This was to fix 
image

Kadrian:
Could you please check all your changes and fix the code formatting?

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const Root = styled("div")<{
+  isOver?: boolean;
+}>`
+  border: ${({ theme, isOver }) =>
+    isOver ? `1px solid ${theme.col.grayMediumLight}` : "none"};
+  position: relative;
+  display: inline-flex;
+  color: ${({ theme, isOver }) =>
+    isOver ? theme.col.red : isOver ? theme.col.black : theme.col.gray};
```
Kadrian, line 20: What does this logic do, can you explain?

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,65 @@
...
+`;
+
+// default time until the hover triggers navigation
+// 600 ms is the default time for a tooltip to appear
+// feels very responsive
+// can be overriden by passing a timeUntilNavigateOverride prop
```
Kadrian, line 26: That note can be removed, right, since there's no  timeUntilNavigateOverride?

frontend/src/js/small-tab-navigation/SmallTabNavigationButton.tsx
```diff
@@ -76,16 +77,19 @@ const SmallTabNavigationButton = forwardRef<HTMLButtonElement, PropsT>(
...
     return (
-      <Button
-        ref={ref}
-        highlightColor={highlightColor}
-        type="button"
-        size={size}
-        selected={isSelected}
-        onClick={onClick}
+      <HoverNavigatable triggerNavigate={onClick}
       >
```
Kadrian, line 81: Still not formatted, is it?

frontend/src/js/small-tab-navigation/SmallTabNavigationButton.tsx
```diff
@@ -10,7 +11,7 @@ const bottomBorderBase = css`
   height: 3px;
   bottom: 0px;
 `;
-const Button = styled("button")<{
+const Button = styled("button") <{
```
Kadrian, line 14: Here as well

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,65 @@
...
+  });
+  return (
+    <Root ref={drop} isOver={isOver}>
+      {children}
+    </Root>
+  );
+};
+
+export const FlexHoverNavigatable = styled(HoverNavigatable)`
+  flex: flex;
```
Kadrian, line 63: What is `flex: flex`?

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,65 @@
...
+    }),
+  });
+  return (
+    <Root ref={drop} isOver={isOver}>
+      {children}
+    </Root>
+  );
+};
+
+export const FlexHoverNavigatable = styled(HoverNavigatable)`
```
Kadrian, line 62: Let's have calling components re-style `HoverNavigatable` if needed

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -0,0 +1,65 @@
...
+interface PropsT {
+  triggerNavigate: () => void;
+  children: ReactNode;
+}
+
+const Root = styled("div")<{
+  isOver?: boolean;
+}>`
+  border: ${({ theme, isOver }) =>
+    isOver ? `1px solid ${theme.col.grayMediumLight}` : "none"};
```
Kadrian, line 16: Using a border for highlight makes the rest of the layout jump. How about using a light gray background color instead?

Kadrian (changes requested):

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
fabian-bizfactory: I have just implemented that using very light gray background. Please let me know what you think

frontend/src/js/standard-query-editor/QueryNode.tsx
```diff
@@ -127,49 +133,51 @@ const QueryNode = ({
...
+        description: node.description,
+        tree: node.tree,
+        tables: node.tables,
+        selects: node.selects,
+
+        additionalInfos: node.additionalInfos,
+        matchingEntries: node.matchingEntries,
+        matchingEntities: node.matchingEntities,
+        dateRange: node.dateRange,
+      }
```
Kadrian, line 147: I added a [CI step for lint & checking format](https://github.com/bakdata/conquery/pull/2856). Could you please merge develop into this branch and fix the formatting issues until the pipeline is green?

frontend/src/js/standard-query-editor/QueryNode.tsx
```diff
@@ -180,37 +187,39 @@ const QueryNode = ({
...
-        excludeFromSecondaryId={node.excludeFromSecondaryId}
-        onDeleteNode={onDeleteNode}
-        onToggleTimestamps={onToggleTimestamps}
-        onToggleSecondaryIdExclude={onToggleSecondaryIdExclude}
-        onExpandClick={expandClick}
-        previousQueryLoading={node.loading}
-        error={node.error}
-      />
-    </Root>
+    <FlexHoverNavigatable triggerNavigate={onClick}>
```
Kadrian, line 190: If we're adding this to `QueryNode`, let's also add it to the concepts in the forms, so to `FormConceptNode`

Kadrian (approved):
Nice, I tested this again, feels good now with the background color change. Code also looks good to me now.

Right now the feature is still missing in `FormConceptNode` – feel free to either add this in here or open a follow up PR

PR #2844, opened 2022-12-06
Fix history date range --> until today by default

Kadrian:


frontend/src/js/api/api.ts
```diff
@@ -398,7 +399,7 @@ export const useGetEntityHistory = () => {
         max: string; // Format like "2020-12-31"
       } = {
         min: "2015-01-01",
```
awildturtok, line 401: you don't have to send a min, is probably safer than wondering why entries are missing

Kadrian: Is `max` optional as well? Is the entire `time` object optional?

awildturtok (approved):

PR #2848, opened 2022-12-08
Fix/frontend env replace

thoniTUB:
@awildturtok 
I found two mismatches with the `replace-env-at-runtime.sh`-script:
- Shebang pointed to `/bin/bash` which is not available in the runner image from the docker file
- `run_e2e_frontend.sh` provided an env-file as argument to the replace-script, whereas the Dockerfile provided the index.html that is to be replaced

The replace-script now takes two arguments. The first one is the index.html and is mandatory, the second  file is an env-file and is optional.

More over i added the `PROXY` build argument to the Dockerfile, so it is possible for yarn to pull dependencies behind a proxy.


frontend/Dockerfile
```diff
@@ -1,5 +1,14 @@
 FROM mhart/alpine-node:16 AS builder
 
+ARG PROXY
+
+# Set up proxy if necessary because yarn does not recognize the corresponding env variables
+RUN if [[ -z $PROXY ]]; then echo "Skip setting up proxy" ; else ( \
```
awildturtok, line 6: you might want to check if yarn looks for lower or uppercase `HTTP_PROXY`. For some reason these both exist and arent universal

frontend/scripts/replace-env-at-runtime.sh
```diff
@@ -1,20 +1,45 @@
...
+
+# Check for one or two arguments or exit 
+if  [ \( "$#" -lt 1 \) -o \( "$#" -gt 2 \) ] 
+then
+    echo "$usage"
+    exit 1
+fi
+
+# Check if env file is provided and readable
+if [ "$#" -eq 2 ]
```
awildturtok, line 20: Why isn't the env already sourced when invoking this file?

awildturtok: You might also want to just use `env` to invoke it

awildturtok (changes requested):

frontend/scripts/replace-env-at-runtime.sh
thoniTUB: but then, i would need to run the following command in a subshell

frontend/Dockerfile
thoniTUB: I just noticed, that using `docker build .` doesn't have problems with the proxy, but using `docker compose up --build` does 🤔

Kadrian (approved):
Tested locally using
```
docker build -t conquery .
docker run -e REACT_APP_API_URL=http://some-conquery-url --name conquery -p 8000:80 conquery
```

The replacement works for me

frontend/Dockerfile
thoniTUB: Habs wieder raus genommen, es geht irgendwie auch ohne wenn man `HTTP_PROXY` etc. als build arg von außen definiert

frontend/scripts/replace-env-at-runtime.sh
thoniTUB: wobei man bei `env` nicht direkt ein env-File angeben kann

awildturtok: `env $(cat ./env) replace-env-at-runtime.sh`?

thoniTUB: Habs probiert, Dafür muss man die env-datei vorher säubern (Kommentare, vmtl auch zeilen umbrüche)

thoniTUB: So würde es gehen: `(source .env.e2e ; ./scripts/replace-env-at-runtime.sh build/index.html.test)`

awildturtok: Fixt du das noch?

awildturtok (approved):

frontend/scripts/replace-env-at-runtime.sh
thoniTUB: Ich hatte es auf dem Server nochmal getestet und muss das sourcen des env-files mit `set -+a` umranden

PR #2849, opened 2022-12-08
added automatic label generation from selected concepts for querynodeeditor

MarcoKorinth:


frontend/src/js/query-node-editor/QueryNodeEditor.tsx
```diff
@@ -159,6 +161,37 @@ const QueryNodeEditor = ({ node, ...props }: QueryNodeEditorPropsT) => {
...
 
+  // Automatic node label generation
+  const formatConceptLabels = (labels: string[]) =>
+    labels
+      .map((label) =>
+        label
+          .split("")
+          .filter((c) => c.match(/[a-zA-Z0-9]/))
+          .join(""),
+      )
```
Kadrian, line 172: you could probably just do `labels.map(label => label.match(...).join(""))`

Kadrian (approved):
Nice one! To keep this logic together, I'd love if you could extract it into an own small hook within QueryNodeEditor like 
```
const {
  autoLabel,
  autoLabelEnabled, 
  setAutoLabelEnabled
} = useAutoLabel({ node, onUpdateLabel });
```

but feel free to merge in any case

frontend/src/js/query-node-editor/QueryNodeEditor.tsx
MarcoKorinth: Yes you're right :)

PR #2870, opened 2023-01-09
Feature/scroll to keep dropzone in view

fabian-bizfactory:


fabian-bizfactory:
Umgesetzt ohne zusätzliche Dependencies

Kadrian (changes requested):
Like explained internally, I'd vote against those changes

PR #2875, opened 2023-01-10
History entity ids iteration

Kadrian:


frontend/src/js/entity-history/LoadHistoryDropzone.tsx
```diff
@@ -46,22 +48,26 @@ export const LoadHistoryDropzone = ({
...
+
+      if (row.length > 2) {
+        loadedEntityStatus[row[1]] = row
+          .slice(2)
+          .filter((str) => str.length > 0)
+          .map((s) => {
+            const opt = s.trim();
+            loadedEntityStatusOptionsRaw.push(opt);
+            return { label: opt, value: opt };
+          });
```
awildturtok, line 70: This section would benefit from an explanation of what `data` is and why you're preferably looking for rows of exactly length 2.

frontend/src/js/entity-history/Navigation.tsx
```diff
@@ -104,17 +104,22 @@ export const Navigation = memo(
...
-
-      const csvString = entityIds
-        .map(idToRow)
-        .map((row) => row.join(";"))
-        .join("\n");
+      const usedStatuses = Object.values(entityIdsStatus).reduce(
+        (longest, el) => (longest.length > el.length ? longest : el),
+        [],
+      );
+      const idToRow = (entityId: EntityId) =>
```
awildturtok, line 111: Why not simply use Papa.unparse for this?

awildturtok (approved):
Neither of these suggestions is a dealbreaker

frontend/src/js/entity-history/LoadHistoryDropzone.tsx
Kadrian: Good point, I grouped the save and load logic + added a comment

frontend/src/js/entity-history/Navigation.tsx
Kadrian: yeah good point, done!

PR #2879, opened 2023-01-12
Feature/auto label

MarcoKorinth:
- delimiter for autolabel elements is now space
- special characters aren't ignored anymore
- elements are sorted in alphabetical order
- autolabel is only applied for nodes with more than one element


Kadrian (approved):
Looks good to me! Got a small suggestion ( non-blocking )

PR #2880, opened 2023-01-12
reword "data set" -> "dataset"

thoniTUB:
Before:  
image

After:
image

### DIFF
--- a/frontend/src/localization/en.json
+++ b/frontend/src/localization/en.json
@@ -82,10 +82,10 @@
     "aConceptList": "a concept list"
   },
   "datasetSelector": {
-    "label": "Data sets",
-    "pleaseChoose": "Choose a data set",
-    "error": "Could not load data sets",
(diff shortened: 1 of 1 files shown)


thoniTUB:
I would prefer `Dataset` which is okay according to https://en.wiktionary.org/wiki/dataset, but other sources say it's less academic

frontend/src/localization/en.json
```diff
@@ -82,10 +82,10 @@
     "aConceptList": "a concept list"
   },
   "datasetSelector": {
-    "label": "Data sets",
```
awildturtok, line 85: Dataset ist ein echtes Wort, warum wird das überhaupt gesplittet?

Kadrian: I agree, let's just use dataset as one word

thoniTUB: 👍

PR #2884, opened 2023-01-16
add Notification Type to Snack Messaging

fabian-bizfactory:


Kadrian:
@fabian-bizfactory please let me know once you've looked through the diff and you think this is ready for review

fabian-bizfactory:
Sorry, about the PR pull. Found a potential issue after creating it

frontend/src/js/snack-message/SnackMessage.tsx
```diff
@@ -41,11 +48,17 @@ const ClearZone = styled("div")`
...
-  const message = useSelector<StateT, string | null>(
-    (state) => state.snackMessage.message,
+  const { message, notificationType } = useSelector<StateT, SnackMessageStateT>(
+    (state) => state.snackMessage,
   );
   const dispatch = useDispatch();
-  const resetMessage = () => dispatch(setMessage({ message: null }));
+  const resetMessage = () =>
+    dispatch(
+      setMessage({
```
Kadrian, line 57: Let's use the `resetMessage` action instead here

frontend/src/js/snack-message/reducer.ts
```diff
@@ -4,12 +4,19 @@ import { Action } from "../app/actions";
 
 import { resetMessage, setMessage } from "./actions";
 
+export enum SnackMessageTypeT {
```
Kadrian, line 7: Ending types with a`T` is something we're trying to slowly move away from, I'd suggest calling it `SnackMessageType`

frontend/src/js/snack-message/SnackMessage.tsx
```diff
@@ -7,13 +7,20 @@ import { useClickOutside } from "../common/helpers/useClickOutside";
 import FaIcon from "../icon/FaIcon";
 
 import { setMessage } from "./actions";
+import { SnackMessageStateT, SnackMessageTypeT } from "./reducer";
 
-const Root = styled("div")`
+const colorLookupTable: { [key in SnackMessageTypeT]: string } = {
```
Kadrian, line 12: Let's avoid calling variables by their "data structure" ("lookup table") and instead describe what they contain. Also `Record` is useful here:
```suggestion
const bgColorByType: Record<SnackMessageType, string> = {
```

frontend/src/js/snack-message/actions.ts
```diff
@@ -1,11 +1,14 @@
...
 
+import { SnackMessageTypeT } from "./reducer";
+
 export type SnackMessageActions = ActionType<
   typeof setMessage | typeof resetMessage
 >;
 
 export const setMessage = createAction("snack-message/SET")<{
   message: string | null;
+  notificationType: SnackMessageTypeT;
```
Kadrian, line 11: how about just calling it `type`? 
```suggestion
  type: SnackMessageType;
```

frontend/src/js/snack-message/SnackMessage.tsx
```diff
@@ -7,13 +7,20 @@ import { useClickOutside } from "../common/helpers/useClickOutside";
...
+  [SnackMessageTypeT.ERROR]: "rgba(0, 0, 0, 0.75)",
+  [SnackMessageTypeT.SUCCESS]: "rgba(12, 100, 39, 0.75)", // #0C6427
```
Kadrian, line 14: Have you tested what it looks like? I think with the green, it's actually too transparent, how about using `0.9`?

Kadrian (changes requested):
Looks good overall, I've got a couple of suggestions, would be happy if you could take a look at those!

frontend/src/js/snack-message/SnackMessage.tsx
fabian-bizfactory: I liked it that way. Took a look at 0.9, you are right, that looks better

Kadrian (approved):
Nice! Looks good to me

PR #2885, opened 2023-01-16
Replace InputMultiSelectDropzone with DropzoneWithFileInput

Kadrian:


awildturtok (approved):
looks good. 

The review shows additional `Unchanged files with check annotations`. Can you take a short look if these are valid warnings or not?

Kadrian:
Yup they're valid warnings. They're appearing now since I added the format and lint check to CI, but they've existed for a long time. Will fix at some point, but safe to ignore for now

PR #2890, opened 2023-01-18
Scroll a little more (20 pixels)

fabian-bizfactory:
Find scrollable parent

New behaviour will scroll instantly into view and will scroll additional 20 pixels smoothly.

Replaces #2870


fabian-bizfactory:
Like discussed, we are using the current version

PR #2903, opened 2023-01-25
Bugfix/autolabel subtitle change

MarcoKorinth:


Kadrian (changes requested):
Hey, sorry for the late review. I tested this and it works, but I think it would be a lot easier if we'd just update the node description on `removeConceptFromNode` in the reducer (i.e. make updating the description completely independent from the "autolabel" feature)

Could you give that another try?

MarcoKorinth:
Yes of course.
You were right, using the removeConceptFromNode hook turned out to be much easier :)

I pushed the new approach to a new branch: https://github.com/ingef/conquery/pull/2924

PR #2904, opened 2023-01-30
also add COLUMN semantics to SECONDARY_ID flagged columns

awildturtok:
@Kadrian this was requested by @MaxVonKnobloch. Can you check if this causes issues with the enity-preview?

Previously, Columns of SecondaryIds are still tossed together, but now their Column is still tagged in that column. ie, secondaryId-Columns may now have multiple COLUMN semantics.

### DIFF
--- a/backend/src/main/java/com/bakdata/conquery/apiv1/query/TableExportQuery.java
+++ b/backend/src/main/java/com/bakdata/conquery/apiv1/query/TableExportQuery.java
@@ -210,6 +210,7 @@ public class TableExportQuery extends Query {
 		infos[0] = ConqueryConstants.DATES_INFO_HISTORY;
 		infos[1] = ConqueryConstants.SOURCE_INFO;
 
+
 		for (Map.Entry<SecondaryIdDescription, Integer> e : secondaryIdPositions.entrySet()) {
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/bakdata/conquery/apiv1/query/TableExportQuery.java
```diff
@@ -242,13 +243,15 @@ private List<ResultInfo> createResultInfos(Map<SecondaryIdDescription, Integer>
...
 
 			final Column column = entry.getKey();
 
+			// 0 Position is date, already covered
 			// SecondaryIds and date columns are pulled to the front, thus already covered.
 			if (position == 0 || column.getSecondaryId() != null) {
+				infos[secondaryIdPositions.get(column.getSecondaryId())].getSemantics()
+																		.add(new SemanticType.ColumnT(column));
 				continue;
 			}
```
thoniTUB, line 256: Ich verstehe leider von den Kommentaren und dem Code her nicht was hier passiert. Wenn man an Position 0, was das date ist, will man dann eine Secondary ID daraus haben?

thoniTUB (approved):

PR #2915, opened 2023-02-01
Fix no newline

Kadrian:
Unclear to me how this happened ?!


thoniTUB:
It might originated from revert on master and following merge into develop I had to do 🤔

thoniTUB (approved):

PR #2931, opened 2023-02-13
Use new Label Request Format in Frontend

fabian-bizfactory:


frontend/mock-api/index.js
```diff
@@ -151,8 +151,14 @@ module.exports = function (app, port) {
               status: "DONE",
               numberOfResults: 5,
               resultUrls: [
-                `/api/results/results.xlsx`,
-                `/api/results/results.csv`,
+                {
+                  label: "XLSX",
+                  url: "http://localhost:8080/api/result/xlsx/mimic-iii-demo.51cd95fd-90b2-4573-aab5-11846126427b.xlsx",
```
Kadrian, line 156: Where do these names + ids come from? `mimic-iii-demo.51cd95fd-90b2-4573-aab5-11846126427b`

frontend/mock-api/stored-queries/25.json
```diff
@@ -2,7 +2,20 @@
...
   "status": "DONE",
   "numberOfResults": 5,
-  "resultUrls": ["/api/results/1234-123123123-123123-1231.csv"],
+  "resultUrls": [
+    {
+      "label": "Alle Dateien",
+      "url": "http://localhost:8080/api/results/all.zip"
+    },
+    {
+      "label": "Bericht.pdf",
```
Kadrian, line 11: So sometimes it's PDF, XLSX, but other times it's an actual label like this? Would it make sense to model this as 
```
{
  label: string;
  url: string;
  ending: string;
}
```

did you discuss something like that?

frontend/src/js/api/types.ts
```diff
@@ -390,12 +390,17 @@ export interface ColumnDescription {
   userConceptLabel: string | null;
 }
 
+export interface ResultUrlsWithLabel {
```
Kadrian, line 393:
```suggestion
export interface ResultUrlWithLabel {
```

frontend/src/js/button/DownloadButton.tsx
```diff
@@ -13,8 +15,31 @@ const Link = styled("a")`
...
+};
+function getFileIcon(label: string): IconName {
+  // Editor Requests
+  if (label in fileTypeToIcon) {
+    return fileTypeToIcon[label];
+  }
+
+  // Forms
+  if (label.includes(".")) {
+    const ext = label.split(".").pop()?.toLocaleUpperCase();
```
Kadrian, line 32: What's happening here? This is getting an ending, right? Can we combine this with [getEnding](https://github.com/ingef/conquery/pull/2931/files#diff-085890ab95def8ca4d7754efd73abd80ab7b62fa279937ecd1ab2d511db5875eR52) here?

Kadrian, line 32: why `toLocaleUpperCase` instead of `toUpperCase`?

frontend/src/js/button/DownloadButton.tsx
```diff
@@ -13,8 +15,31 @@ const Link = styled("a")`
...
+    if (!ext) return "file-download";
```
Kadrian, line 33: Is this check really required here?

frontend/src/js/button/DownloadButton.tsx
```diff
@@ -13,8 +15,31 @@ const Link = styled("a")`
...
+    if (ext in fileTypeToIcon) {
+      return fileTypeToIcon[ext];
+    }
+  }
+  return "file-download";
+}
+
 interface Props extends Omit<IconButtonPropsT, "icon" | "onClick"> {
-  url: string;
+  url: ResultUrlsWithLabel;
```
Kadrian, line 42: Let's please avoid `url.url` anywhere. Here you could do
```suggestion
  resultUrl: ResultUrlWithLabel;
```

frontend/src/js/button/PreviewButton.tsx
```diff
@@ -29,7 +29,10 @@ const PreviewButton: FC<PropsT> = ({ url, columns, ...restProps }) => {
   return (
     <Button
       onClick={async () => {
-        await loadPreviewData(getAuthorizedUrl(url), columns);
+        await loadPreviewData(
+          { ...url, url: getAuthorizedUrl(url.url) },
```
Kadrian, line 33: I don't think we should make this change. Preview doesn't need the `label`, does it? Maybe just keep the `url` prop a `string`?

frontend/src/js/button/QueryResultHistoryButton.tsx
```diff
@@ -35,7 +35,11 @@ export const QueryResultHistoryButton = ({ url, label, columns }: PropsT) => {
       icon={isLoading ? "spinner" : "list-ul"}
       frame
       onClick={async () => {
-        await newHistorySession(getAuthorizedUrl(url), columns, label);
+        await newHistorySession(
+          { ...url, url: getAuthorizedUrl(url.url) },
```
Kadrian, line 39: Same here

frontend/src/js/entity-history/actions.ts
```diff
@@ -73,12 +74,12 @@ export const loadHistoryData = createAsyncAction(
 )<
   void,
   {
-    currentEntityCsvUrl: string;
+    currentEntityCsvUrl: ResultUrlsWithLabel;
```
Kadrian, line 77: I think this should remain a string as well (unless you see some benefit of adding the label information here? --> please explain)

frontend/src/js/entity-history/actions.ts
```diff
@@ -195,13 +200,13 @@ export function useUpdateHistorySession() {
             defaultEntityHistoryParams.sources,
           );
 
-        const csvUrl = resultUrls.find((url) => url.endsWith("csv"));
+        const csvUrl = resultUrls.find((url) => url.label === "CSV");
```
Kadrian, line 203: Would this even work? Would there always be a `label` === "CSV" for every editor query?

frontend/src/js/entity-history/reducer.ts
```diff
@@ -39,13 +40,13 @@ export type EntityHistoryStateT = {
...
   columnDescriptions: ColumnDescription[];
-  resultUrls: string[];
+  resultUrls: ResultUrlsWithLabel[];
   label: string;
   entityIds: EntityId[];
   currentEntityUniqueSources: string[];
   currentEntityId: EntityId | null;
   currentEntityData: EntityEvent[];
-  currentEntityCsvUrl: string;
+  currentEntityCsvUrl: ResultUrlsWithLabel | null;
```
Kadrian, line 49: See comment above

frontend/src/js/pane/TabNavigation.tsx
```diff
@@ -98,4 +98,4 @@ const TabNavigation: FC<PropsT> = ({
   );
 };
 
-export default TabNavigation;
\ No newline at end of file
+export default TabNavigation;
```
Kadrian, line 101: We made this change in `develop` already, could you try to merge develop in?

frontend/src/js/preview/reducer.ts
```diff
@@ -1,14 +1,14 @@
...
+import type { ColumnDescription, ResultUrlsWithLabel } from "../api/types";
 import { Action } from "../app/actions";
 
 import { openPreview, closePreview, loadCSVForPreview } from "./actions";
 
 export type PreviewStateT = {
   isOpen: boolean;
   isLoading: boolean;
-  dataLoadedForResultUrl: string | null;
+  dataLoadedForResultUrl: ResultUrlsWithLabel | null;
```
Kadrian, line 11: This should remain a string as well I think

frontend/src/js/query-runner/DownloadResultsDropdownButton.tsx
```diff
@@ -43,56 +44,76 @@ const Separator = styled("div")`
...
+    const labelResult = resultUrls.find(
+      (url) => url.label === fileChoice.label,
+    );
+    if (labelResult) {
+      return labelResult;
+    }
+    return (
+      resultUrls.find((url) => getEnding(url.url) === fileChoice.ending) ||
+      resultUrls[0]
+    );
```
Kadrian, line 100: This looks very similar to the code above for `getInitialEndingChoice`, could you try to combine both? Maybe extract a function and simplify the code like
```
(
  resultUrls.find((url) => url.label === fileChoice.label) ||
  resultUrls.find((url) => getEnding(url.url) === fileChoice.ending) ||
  resultUrls[0]
);
```

frontend/src/js/user/userSettings.ts
```diff
@@ -4,6 +4,7 @@ const localStorage: Storage = window.localStorage;
 interface UserSettings {
   arePreviousQueriesFoldersOpen: boolean;
   preferredDownloadFormat?: string; // Usually CSV or XLSX
+  preferredDownloadLabel?: string; // Label of the prefered Download format (e.g. "Alle Dateien")
```
Kadrian, line 7:
```suggestion
  preferredDownloadLabel?: string; // label of the preferred download format (e.g. "All files")
```

frontend/src/js/user/userSettings.ts
```diff
@@ -4,6 +4,7 @@ const localStorage: Storage = window.localStorage;
 interface UserSettings {
   arePreviousQueriesFoldersOpen: boolean;
   preferredDownloadFormat?: string; // Usually CSV or XLSX
```
Kadrian, line 6: Maybe we could even rename this ?
```suggestion
  preferredDownloadEnding?: string; // Usually CSV or XLSX
```

frontend/src/js/query-runner/DownloadResultsDropdownButton.tsx
```diff
@@ -43,56 +44,76 @@ const Separator = styled("div")`
...
+  }, [resultUrls, fileChoice]);
 
   const dropdown = useMemo(() => {
     return (
       <List>
         {resultUrls.map((url) => {
```
Kadrian, line 106: If you destructure like this you can avoid url.url
```suggestion
        {resultUrls.map(({ url, label }) => {
```

frontend/src/js/query-runner/QueryResults.tsx
```diff
@@ -49,7 +49,7 @@ const QueryResults: FC<PropsT> = ({
   queryType,
 }) => {
   const { t } = useTranslation();
-  const csvUrl = resultUrls.find((url) => url.endsWith("csv"));
+  const csvUrl = resultUrls.find((ru) => ru.url.endsWith("csv"));
```
Kadrian, line 52: Destructuring here would help as well

frontend/src/js/query-runner/DownloadResultsDropdownButton.tsx
```diff
@@ -105,7 +126,7 @@ const DownloadResultsDropdownButton = ({
       {!tiny && (
         <>
           <SxDownloadButton bgHover url={urlChoice}>
-            {endingChoice}
+            {urlChoice.label}
```
Kadrian, line 129: Did you test what it looks like with longer labels? Is there a max length we should restrict this to?

Kadrian (changes requested):
Going into a good direction! I left some comments.

frontend/src/js/query-runner/DownloadResultsDropdownButton.tsx
fabian-bizfactory: image
The content will expand to the left. I think that looks acceptable, since I believe extremely long labels will be rare. If you think that it is necessary, I can write a function that truncates the output

frontend/mock-api/stored-queries/25.json
fabian-bizfactory: I don't think that it we should do it like that, since we are already transferring the "real" ending as the url ending. So we would only be transferring redundant data.

frontend/src/js/entity-history/actions.ts
fabian-bizfactory: I think it should, but reverting this anyways to the `.endswith("csv")` method. I believe that should be more robust

Kadrian:
There seems to be something wrong with the commits and changes here @fabian-bizfactory

fabian-bizfactory:
> There seems to be something wrong with the commits and changes here @fabian-bizfactory

@Kadrian I believe that is what it should look like, since this is a request to merge into the other feature branch to combine the backend and frontend changes. So this will also update the [feature/result-assets](https://github.com/ingef/conquery/tree/feature/result-assets) to develop.

thoniTUB:
@fabian-bizfactory @Kadrian I think this branch was not based on [feature/result-assets](https://github.com/ingef/conquery/tree/feature/result-assets) and now merged into it. I'll rebase feature/result-assets onto develop, that might shrink the number of commits

thoniTUB:
@fabian-bizfactory i'll rebase this branch onto [feature/result-assets](https://github.com/ingef/conquery/tree/feature/result-assets), so that @Kadrian has a clean review. Keep in mind that i'll force push this branch and your local git-repo will complain

thoniTUB:
@Kadrian should be good to review now.

frontend/mock-api/stored-queries/25.json
Kadrian: Okay, so are these three examples common for a `stored query`? If I understood correctly, the editor would always yield a CSV, right? And a form query result would yield these pdf / xlsx / other, right?

So we should use the CSV / XLSX examples here, no?

frontend/src/js/query-runner/DownloadResultsDropdownButton.tsx
```diff
@@ -43,56 +44,82 @@ const Separator = styled("div")`
 // might need to adjust this when adding more content.
 const dropdownOffset: [number, number] = [-37, 8]; // [skidding, distance] / default [0, 10]
```
Kadrian, line 45: Now that the content in the dropdown is changed, this skidding / distance setting needs to change as well: the skidding helped align the right edge of the dropdown with the button element. Now with longer content, it's no longer aligned and sticks to the right edge of the screen.

To re-align things properly, here's a PR https://github.com/ingef/conquery/pull/2944

frontend/src/js/query-runner/DownloadResultsDropdownButton.tsx
Kadrian: The overlay content itself will be fine, but I'm asking about the button label itself. 

Because this is what it will look like

Screenshot 2023-02-20 at 12 19 56

> I believe extremely long labels will be rare

Could you discuss with the team members what exactly they will look like?

If longer than 40 characters could happen, I'd say we have to add truncation

Kadrian (changes requested):
Alright, looks a lot better already. Review round 2 done. Some last things remaining

frontend/src/js/query-runner/DownloadResultsDropdownButton.tsx
fabian-bizfactory: That makes sense, I just added the truncation for it. Will truncate after 40 chars to 37 plus ...

fabian-bizfactory:
Will merge the updates from develop in a few minutes, then it will be good to review again

Kadrian:
@fabian-bizfactory when the base branch `feature/result-assets` isn't up to date, this isn't easy to review

thoniTUB:
It look like the test captured an outdated token, which should not happen. I'm looking into that

thoniTUB:
I rerun the test. I suppose there was a timing glitch. Normally, the test has one minute to complete, but usually take one some millis.

fabian-bizfactory:
Thanks @thoniTUB!
@Kadrian should be good to review now

thoniTUB:
wouhpsie, I was mistaken, the test had only one second to complete. I'll relax that on the base branch

frontend/mock-api/index.js
Kadrian: @fabian-bizfactory ?

Kadrian (approved):
I'm still a little confused about the mock examples, where we're using a name like `/mimic-iii-demo.51cd95fd-90b2-4573-aab5-11846126427b.xlsx` in one place and `/1234-123123123-123123-1231.xlsx` in another. You resolved the conversation and forgot to answer the question, see above. Right now we're also missing a mock api example for a name that would get truncated. But overall, not a big deal.

The code looks good to me now, and it works for me locally as well, thanks for adding the truncation.

thoniTUB:
@Kadrian @fabian-bizfactory i removed the mimic-iii part and added a dicing for a result label which is too long

frontend/mock-api/index.js
thoniTUB: they are propably from the mimic III tutorial dataset

fabian-bizfactory: Yes, they are from the tutorial dataset. The ids are generated from the backend.

PR #2944, opened 2023-02-20
Fix tooltip alignment with popperOptions using padding instead of skidding

Kadrian:
See https://github.com/ingef/conquery/pull/2931#discussion_r1111812959


fabian-bizfactory (approved):
Looks good

PR #2964, opened 2023-03-03
Feature/synchronous add deletion dataset

thoniTUB:
The manager node waits until it has add/delete confirmation from all shards/workers.

Came up with new e2e tests for the admin-ui


thoniTUB:
Closed because the approch is cumbersome. Use http instead of mina

PR #2965, opened 2023-03-03
use different ids for dataset in e2e test to avoid timing bux

thoniTUB:
This is just a quick workaround for a timing-bug that occurs when a dataset with the same dataset-Id is created, deleted, created.

An actual fix is here: https://github.com/ingef/conquery/pull/2964


thoniTUB:
@awildturtok I merge this little workaround now, so that the test is green for the next iteration

PR #2969, opened 2023-03-07
fix field transform to not return null when value=0

fabian-bizfactory:


frontend/src/js/external-forms/transformQueryToApi.ts
```diff
@@ -46,7 +46,7 @@ function transformFieldToApiEntries(
     case "TEXTAREA":
     case "STRING":
     case "NUMBER":
-      return [[fieldConfig.name, formValue || null]];
+      return [[fieldConfig.name, formValue !== undefined ? formValue : null]];
```
Kadrian, line 49:
```suggestion
      return [[fieldConfig.name, formValue ?? null]];
```

Kadrian (approved):
I think double question mark would be what we want here, feel free to adjust if you like

PR #2971, opened 2023-03-08
add root concept info in Tooltip

fabian-bizfactory:


fabian-bizfactory:
> @fabian-bizfactory have you checked the diff / the warnings? Is this ready to review?

Sorry, I missed the lint errors. I just fixed them. It is ready to review now

Kadrian:
@fabian-bizfactory could you merge in develop? Please make sure to run `yarn` locally. Icons need to be individually imported (see here for examples https://github.com/ingef/conquery/pull/2978/files )

fabian-bizfactory:
@Kadrian FYI I had to remove 
```js
define: {
      __BUILD_TIMESTAMP__: new Date()
        .toISOString()
        .split(".")[0]
        .split("T")
        .join(" "),
    },
```
from `vite.config.ts` since that broke my dev server. I did not commit that change though.

Kadrian:
could you provide details on

>  that broke my dev server.

fabian-bizfactory:
> could you provide details on
> 
> > broke my dev server.

On firefox (110.0.1 (64-bit)) I get the error: 
`Uncaught SyntaxError: "0"-prefixed octal literals are deprecated; use the "0o" prefix instead`

On Chrome:
`Uncaught SyntaxError: Unexpected number`

Both are errors in `env.mjs:16` from vite

frontend/src/js/tooltip/AdditionalInfoHoverable.tsx
```diff
@@ -13,32 +14,39 @@ const Root = styled("div")`
...
+  parent?: string,
+): AdditionalInfosType => ({
   label: node.label,
   description: node.description,
   isFolder: !!node.children && node.children.length > 0,
   matchingEntries: node.matchingEntries,
   matchingEntities: node.matchingEntities,
   dateRange: node.dateRange,
   infos: node.additionalInfos,
+  parent: parent ? getConceptById(parent)?.label : null,
```
Kadrian, line 28: What's the performance impact from that? Like this, every time someone hovers, we'll do a lookup and that's probably not instant.

I'd suggest to 
a) at least pass the root id a second param as well for optimized lookup, see implementation of the function
b) better even: do the optimized root lookup when the ConceptTreeNode renders and pass the root `label` directly, instead of the root id, maybe even re-use the looked up root in `createQueryElement`

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -80,6 +83,10 @@ const Infos = styled("div")`
   overflow-x: auto;
 `;
 
+const IndentRoot = styled("div")<{ depth: number }>`
+  padding-left: ${({ depth }) => depth * 15 + "px"};
```
Kadrian, line 87: Since you're only using depth=1, could you please just remove depth and just do
```suggestion
  padding-left: 15px;
```

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -141,8 +146,14 @@ const Tooltip = () => {
     matchingEntries,
     matchingEntities,
     dateRange,
+    parent,
   } = additionalInfos;
 
+  const hasChild = useMemo(() => {
+    return parent !== label && parent;
+  }, [parent, label]);
+
```
Kadrian, line 155: Do you think a useMemo is necessary here? Why?

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -141,8 +146,14 @@ const Tooltip = () => {
...
+  if (!displayTooltip) return <ActivateTooltip />;
```
Kadrian, line 156: For early returns, please give it some space
```suggestion
  if (!displayTooltip) return <ActivateTooltip />;
  
```

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -164,9 +175,21 @@ const Tooltip = () => {
         />
         <Head>
           <PinnedLabel>
-            <TypeIcon icon={isFolder ? faFolder : faMinus} />
+            <TypeIcon
+              icon={
+                isFolder || parent !== label
+                  ? hasChild
```
Kadrian, line 181: A little hard to understand: `hasChild` is also checking `parent !== label`,  right? Could you make this a little easier to read / understand?

frontend/src/js/tooltip/reducer.ts
```diff
@@ -22,6 +22,7 @@ export type AdditionalInfosType = {
   matchingEntities: number | null;
   dateRange?: DateRangeT;
   infos?: InfoType[];
+  parent?: string | null;
```
Kadrian, line 25: any reason to allow null **and** undefined here? How about just
```suggestion
  parent?: string
```

Kadrian (changes requested):

Kadrian:
On the vite define errors
- which node version are you using? (should be 18)
- could you try to 
  - delete node_modules
  - run `yarn`
  - run `yarn dev`

fabian-bizfactory:
> On the vite define errors
> 
>     * which node version are you using? (should be 18)
> 
>     * could you try to
>       
>       * delete node_modules
>       * run `yarn`
>       * run `yarn dev`

I was on node 19, but downgraded to 18(18.5.0) just now, same issue. I have tried the second process as well, but that did not fix the issues.

Kadrian:
Fix for the build timestamp is merged: https://github.com/ingef/conquery/pull/2991

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -164,9 +171,23 @@ const Tooltip = () => {
         />
         <Head>
           <PinnedLabel>
-            <TypeIcon icon={isFolder ? faFolder : faMinus} />
+            <TypeIcon
+              icon={
+                // Parent is not element tooltipped or tooltip is a folder
+                isFolder || parent !== label
```
Kadrian, line 177: Hmm, I still have a hard time understanding this, is there a case for something that's `!isFolder` but `parent !== label`?

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -177,6 +198,17 @@ const Tooltip = () => {
               />
             )}
           </PinnedLabel>
+          {parent !== label && parent ? (
+            <IndentRoot>
+              <PinnedLabel>
+                <TypeIcon icon={faCaretRight} />
```
Kadrian, line 204: Hmm, the caret suggests that you can click this to open, no? So I don't think it makes sense to show it here, how about just removing it?

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -177,6 +198,17 @@ const Tooltip = () => {
               />
             )}
           </PinnedLabel>
+          {parent !== label && parent ? (
+            <IndentRoot>
```
Kadrian, line 202: This looks pretty squished right underneath the parent label. Could you give this some space to top and bottom? Maybe something like `margin: 3px 0 8px`?

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -55,7 +57,7 @@ const TackIconButton = styled(IconButton)`
   margin-left: 5px;
 `;
 const TypeIcon = styled(StyledFaIcon)`
-  margin-right: 10px;
+  margin-right: 5px;
```
Kadrian, line 60: Nice, I think that's more similar to what it looks like in the concept trees as well, exact would probably be
```suggestion
  margin-right: 6px;
```

frontend/src/js/tooltip/AdditionalInfoHoverable.tsx
```diff
@@ -27,18 +27,24 @@ const AdditionalInfoHoverable = ({
...
 
   const onDisplayAdditionalInfos = () => {
     if (!node.additionalInfos && isEmpty(node.matchingEntries)) return;
-
     dispatch(
-      displayAdditionalInfos({ additionalInfos: getAdditionalInfos(node) }),
+      displayAdditionalInfos({
+        additionalInfos: {
+          ...getAdditionalInfos(node),
+          parent,
```
Kadrian, line 45: Now adding it outside of the `getAdditionalInfos` function is a little confusing when `getAdditionalInfos` returns an `AdditionalInfosType` which includes `parent?: string`, no?

frontend/src/js/tooltip/AdditionalInfoHoverable.tsx
```diff
@@ -27,18 +27,24 @@ const AdditionalInfoHoverable = ({
   node,
   className,
   children,
+  parent,
 }: {
   children: ReactNode;
   node: ConceptT;
   className?: string;
+  parent?: string;
```
Kadrian, line 35: And this is the parentLabel now, right? So how about calling it
```suggestion
  parentLabel?: string;
```

Kadrian (changes requested):
Feels like we're almost there

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -164,9 +171,23 @@ const Tooltip = () => {
...
         <Head>
           <PinnedLabel>
-            <TypeIcon icon={isFolder ? faFolder : faMinus} />
+            <TypeIcon
+              icon={
+                // Parent is not element tooltipped or tooltip is a folder
+                isFolder || parent !== label
+                  ? // If the parent exists we show the folder icon
+                    parent
+                    ? faFolderOpen
```
Kadrian, line 180: Hmm. So when parent is the same as label, we still show the `faFolderOpen`?  I don't think this makes sense. I'd vote to never show a `faFolderOpen`, but instead show a faFolder `regular` icon for the  concept tree "category" folders, to match the icons displayed in the tree. The attribute to check would be `const isTreeFolder = !parent.detailsAvailable`

frontend/src/js/tooltip/Tooltip.tsx
fabian-bizfactory: I just tried again to leave either out, but that creates the following issue:
- if the child is a Concept tree node, then the parent will also be rendered with `fa-minus` (if we leave `parent !== label`
- if we leave `is-folder` then the "big" folders without a parent folder will be rendered with `fa-minus`

My goal here was to mimic the icons from the concept tree.

frontend/src/js/tooltip/AdditionalInfoHoverable.tsx
```diff
@@ -13,32 +13,41 @@ const Root = styled("div")`
...
 
   const onDisplayAdditionalInfos = () => {
     if (!node.additionalInfos && isEmpty(node.matchingEntries)) return;
-
     dispatch(
-      displayAdditionalInfos({ additionalInfos: getAdditionalInfos(node) }),
+      displayAdditionalInfos({
+        additionalInfos: {
+          ...getAdditionalInfos(node, parent),
+        },
```
Kadrian, line 49: No need for the spread anymore, right?

frontend/src/js/tooltip/reducer.ts
```diff
@@ -22,6 +22,7 @@ export type AdditionalInfosType = {
   matchingEntities: number | null;
   dateRange?: DateRangeT;
   infos?: InfoType[];
+  parent?: ConceptT | null;
```
Kadrian, line 25: Let's please avoid using both `undefined | null`

frontend/src/js/tooltip/Tooltip.tsx
Kadrian: Yeah we should definitely mimic the icons from the concept tree. Maybe this gets clearer extracted to a function and using some if statements instead of nested ternary?

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -164,9 +171,23 @@ const Tooltip = () => {
...
-            <TypeIcon icon={isFolder ? faFolder : faMinus} />
+            <TypeIcon
+              icon={
+                // parent?.label is not element tooltipped or tooltip is a folder
+                isFolder || parent?.label !== label
+                  ? // If the parent is a struct folder, we show the regular folder icon
+                    !parent?.detailsAvailable
+                    ? faFolderRegular
+                    : faFolder
+                  : faMinus
```
Kadrian, line 182: In order to make this more readable, can I ask you to
- extract this to a function
- avoid nested ternary 
- avoid comments that say the same thing that the code says

frontend/src/js/concept-trees/ConceptTreeNodeTextContainer.tsx
```diff
@@ -89,7 +92,12 @@ const ConceptTreeNodeTextContainer: FC<PropsT> = ({
   });
 
   return (
-    <AdditionalInfoHoverable node={node}>
+    <AdditionalInfoHoverable
+      node={node}
+      parent={
+        parentId ? getConceptById(parentId, parentId) ?? undefined : undefined
```
Kadrian, line 98: I guess this is an indication that we should refactor `getConceptById` to return `undefined` instead of `null`, but no need to do that now, I can follow up with a PR here

frontend/src/js/tooltip/Tooltip.tsx
fabian-bizfactory: @Kadrian would this be readable enough for you? I extracted it to a function and added the extracted variables.
```ts
  const chooseIcon = () => {
    const hasParent = parent?.label !== label;
    if(isFolder || hasParent) {
      const isStructNode = !parent?.detailsAvailable;
      if(isStructNode) {
        return faFolderRegular
      }
      return faFolder
    }
    return faMinus
  }
```

Kadrian: Hmm, that's a good start. But what's confusing to me there is you'd be saying: "if it does **not** have a parent, return a faMinus" But faMinus concepts usually do have a parent. Maybe `hasParent` is not a good name here? Also: will this benefit from early returns?

Kadrian: definitely the second one!

fabian-bizfactory: Just implemented that one, should hopefully be good now. Check are failing due to java codeql

Kadrian (approved):

frontend/src/js/tooltip/Tooltip.tsx
thoniTUB: The Java codeql Problem seems to be a spurious Network Glitch durding dependency download. For now we don't have fix for this

PR #2994, opened 2023-03-20
removed force on click and added check if delete buttons are detached

FabianBinder21:


cypress/integration/backend-admin-ui/test_1_datasets.js
```diff
@@ -31,7 +31,10 @@ context("Admin UI Datasets", () => {
         before(() => { visitAdminUI('datasets'); });
 
         it("Can delete the test dataset", () => {
-            cy.get(`[data-test-id="delete-btn-${testDSID}"]`).click({force: true});
+            cy.get(`[data-test-id="delete-btn-${testDSID}"]`)
+                .should($el => {
+                    expect(Cypress.dom.isDetached($el)).to.eq(false)
+                }).click();
```
Kadrian, line 37: Not very obvious to me why it would be detached at this point? Could you add an explanation? Maybe extract a small function?

FabianBinder21: You are right, at this point it's not detached. I removed the check for all elements that shouldn't be detached. The check is still needed for some elements. After the successful upload of the test table and concept a page reload happens. After the reload it appears that the delete buttons are detached.

Kadrian: Could you add the explanation to the code?

FabianBinder21: I thought it would re-attach the element, what is wrong. After talking with Max we figured out that the loading time of bootstrap could be the issue. A strong indication of this is that the error only occurs very inconsistently. We decided to close this pr and Max is going to create a new issue regarding bootstrap loading.

PR #3006, opened 2023-03-28
Refactor additional info hover

Kadrian:
- Fix root display for hovering query nodes and concept entries
- Refactor global tree getter
- Extract some smaller components from Tooltip
- Extract a shared node icon helper to always get the same icon for a concept

@fabian-bizfactory FYI, I also added some comments to explain

### DIFF
--- a/frontend/src/js/concept-trees/globalTreeStoreHelper.ts
+++ b/frontend/src/js/concept-trees/globalTreeStoreHelper.ts
@@ -49,9 +49,9 @@ export function setTree(
 export function getConceptById(
   conceptId: ConceptIdT,
   rootId?: ConceptIdT,
-): ConceptT | null {
+): ConceptT | undefined {
   if (rootId) {
-    return window.conceptTrees[rootId][conceptId] || null;
+    return window.conceptTrees[rootId]?.[conceptId];
   }
 
   const conceptTreeRootIds: ConceptIdT[] = Object.keys(window.conceptTrees);
@@ -62,7 +62,7 @@ export function getConceptById(
     if (concept) return concept;
   }
 
-  return null;
+  return undefined;
 }
 
 //
--- a/frontend/src/js/model/node.ts
+++ b/frontend/src/js/model/node.ts
@@ -1,3 +1,13 @@
+import {
+  faFolder as faFolderRegular,
+  faFolderOpen as faFolderOpenRegular,
+} from "@fortawesome/free-regular-svg-icons";
+import {
+  faFolder,
+  faFolderOpen,
+  faMinus,
+} from "@fortawesome/free-solid-svg-icons";
+
 import { ConceptElementT, ConceptT } from "../api/types";
 import { DNDType } from "../common/constants/dndTypes";
 import type {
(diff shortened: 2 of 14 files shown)


frontend/src/js/model/node.ts
```diff
@@ -96,3 +107,23 @@ export function nodeIsAllowlisted(
 export function nodeIsElement(node: ConceptT): node is ConceptElementT {
   return "tables" in node;
 }
+
+export function getNodeIcon(
```
Kadrian, line 111: helper to make sure we're using the same logic for displaying the right icon for a node

frontend/src/js/tooltip/AdditionalInfoHoverable.tsx
```diff
@@ -15,36 +16,39 @@ const Root = styled("div")`
...
+  root: ConceptT;
 }) => {
   const dispatch = useDispatch();
 
   const onDisplayAdditionalInfos = () => {
     if (!node.additionalInfos && isEmpty(node.matchingEntries)) return;
     dispatch(
       displayAdditionalInfos({
-        additionalInfos: getAdditionalInfos(node, parent),
+        additionalInfos: getAdditionalInfos(node, root),
```
Kadrian, line 51: calling it `root`, not `parent`, because strictly saying, it's not really the parent

frontend/src/js/tooltip/AdditionalInfoHoverable.tsx
```diff
@@ -15,36 +16,39 @@ const Root = styled("div")`
...
   matchingEntries: node.matchingEntries,
   matchingEntities: node.matchingEntities,
   dateRange: node.dateRange,
   infos: node.additionalInfos,
-  parent: parent,
+  icon: getNodeIcon(node, {
+    isStructNode: !root?.detailsAvailable,
+  }),
+  rootLabel: root?.label,
+  rootIcon: root ? getNodeIcon(root) : undefined,
```
Kadrian, line 31: Computing the icon already in the action payload, allows to remove `isFolder` and the only thing left that's required from `root` is its `label` then. That avoids passing a whole node (potentially a very large object including all children, not so great with redux state)

frontend/src/js/standard-query-editor/QueryNode.tsx
```diff
@@ -245,11 +246,12 @@ const QueryNode = ({
...
   if (nodeIsConceptQueryNode(node)) {
-    const conceptById = getConceptById(node.ids[0]);
+    const conceptById = getConceptById(node.ids[0], node.tree);
+    const root = getConceptById(node.tree, node.tree);
 
-    if (conceptById) {
+    if (conceptById && root) {
       return (
-        <AdditionalInfoHoverable node={conceptById}>
+        <AdditionalInfoHoverable node={conceptById} root={root}>
```
Kadrian, line 254: Have to add the root info here to make QueryNodes hoverable too

frontend/src/js/concept-trees/globalTreeStoreHelper.ts
```diff
@@ -49,9 +49,9 @@ export function setTree(
 export function getConceptById(
   conceptId: ConceptIdT,
   rootId?: ConceptIdT,
-): ConceptT | null {
+): ConceptT | undefined {
```
Kadrian, line 52: Refactored the global helper to return undefined, makes for easier handling of the return value in other places

thoniTUB: Is `ConceptT?` not supported by TS or is it something different in this context?

Kadrian: `ConceptT?` is not valid syntax as far as I know

fabian-bizfactory (commented):
Thanks for the xplanations, looks good!

PR #3007, opened 2023-03-28
splitted deletion of test concept and table into seperate test cases …

FabianBinder21:
…and added delay

### DIFF
--- a/cypress/integration/backend-admin-ui/test_2_dataset.js
+++ b/cypress/integration/backend-admin-ui/test_2_dataset.js
@@ -61,18 +61,14 @@ context("Admin UI Single Dataset", () => {
         });
     });
 
-    describe("Can delete test concept", () => {
-        before(() => { visitAdminUI(`datasets/${testDSID}`); });
+    describe("Can delete test concept and table", () => {
+        beforeEach(() => visitAdminUI(`datasets/${testDSID}`));
 
         it("Can delete test concept", () => {
             cy.get('[data-test-id="accordion-Concepts"]').click();
             cy.get(`[data-test-id="delete-btn-concept-${testDSID}.concept1"]`).click({force: true});
             cy.get(`[data-test-id="delete-btn-${testDSID}.concept1"]`).should('not.exist');
         });
-    });
-
-    describe("Can delete test table", () => {
-        before(() => { visitAdminUI(`datasets/${testDSID}`); });
 
         it("Can delete test table", () => {
             cy.get('[data-test-id="accordion-Tables"]').click();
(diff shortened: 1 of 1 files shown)


FabianBinder21:
@thoniTUB we will restart the e2e-tests multiple times to ensure that the fix worked. After that i will assign you as the reviewer.

FabianBinder21:
@thoniTUB I have started the e2e job multiple times without any failures. Our thought process was, that the docker container performance varies and after the page reload triggered by previously deleting the test concept, the test runs before the page reload has finished. Therefore we added a delay

cypress/integration/backend-admin-ui/test_2_dataset.js
```diff
@@ -61,20 +61,21 @@ context("Admin UI Single Dataset", () => {
...
         it("Can delete test concept", () => {
             cy.get('[data-test-id="accordion-Concepts"]').click();
             cy.get(`[data-test-id="delete-btn-concept-${testDSID}.concept1"]`).click({force: true});
             cy.get(`[data-test-id="delete-btn-${testDSID}.concept1"]`).should('not.exist');
         });
+    });
 
+    describe("Can delete test table", () => {
         it("Can delete test table", () => {
+            cy.wait(100);
```
thoniTUB, line 74: Does this wait decides whether the test passes or not?
It looks tricky.
If possible i propose to call `visitAdminUI(`datasets/${testDSID}`);` instead

FabianBinder21: It should make a difference. Locally there is no change if the wait is set or not, it always passes.
We thought about using `visitAdminUI` but I was unsure, if it actually reloads the page if the same url is given.
But after reading the documentation (https://docs.cypress.io/api/commands/visit) it seems like it should always reload.
I will change it to `visitAdminUI`.

cypress/integration/backend-admin-ui/test_2_dataset.js
```diff
@@ -61,20 +61,24 @@ context("Admin UI Single Dataset", () => {
...
+        before(() => { visitAdminUI(`datasets/${testDSID}`); });
 
         it("Can delete test concept", () => {
             cy.get('[data-test-id="accordion-Concepts"]').click();
             cy.get(`[data-test-id="delete-btn-concept-${testDSID}.concept1"]`).click({force: true});
             cy.get(`[data-test-id="delete-btn-${testDSID}.concept1"]`).should('not.exist');
         });
+    });
+
+    describe("Can delete test table", () => {
```
thoniTUB, line 74: If you know put both `it` back in one `describe` it should have the same effect, doesn't it?

FabianBinder21: Yes, but only if `visitAdminUI` is called after the test concept got deleted. Therefore i changed `before` to 'beforeEach` to make sure the page does reload before the deletion of the test table and combined both tests under one `describe` again.
If you got nothing more to add i would restart the e2e job multiple times again to make sure it works and merge the pr afterwards.

PR #3014, opened 2023-04-03
Use suggested min timestamp for history

Kadrian:


frontend/src/js/api/types.ts
```diff
@@ -533,6 +533,7 @@ export interface HistorySources {
 export type GetEntityHistoryDefaultParamsResponse = HistorySources & {
   searchConcept: string | null; // concept id
   searchFilters?: string[]; // allowlisted filter ids within the searchConcept
+  observationPeriodMin: string; // yyyy-MM-dd
```
Kadrian, line 536: Is this always defined?

awildturtok: yep

awildturtok (approved):

PR #3016, opened 2023-04-03
Add api types for new time stratified infos

Kadrian:


frontend/src/js/api/types.ts
```diff
@@ -542,10 +542,39 @@ export interface EntityInfo {
...
+}
+
+export interface TimeStratifiedInfo {
+  label: string;
+  description: string | null;
+  columns: {
+    label: string; // Matches `label` with `year.values` and `year.quarters[].values`
+    defaultLabel: string; // Probably not used by us
+    description: string | null;
+    type: ColumnDescriptionKind; // Probably not used by us
```
awildturtok, line 567: For money this would be relevant to append the proper suffix

awildturtok (approved):

frontend/src/js/api/types.ts
Kadrian: good point, updated

PR #3020, opened 2023-04-04
Persist form values when switching form

Kadrian:


frontend/src/js/external-forms/FormConfigLoader.tsx
```diff
@@ -124,7 +124,7 @@ const FormConfigLoader: FC<Props> = ({
         // from string, e.g. 'next'
         // to SelectValueT, e.g. { value: 'next', label: 'Next' }
         const field = collectAllFormFields(formConfig.fields).find(
-          (f) => f.type !== "GROUP" && f.name === fieldname,
+          (f): f is Field | Tabs => f.type !== "GROUP" && f.name === fieldname,
```
awildturtok, line 127: I'm a  bit mystified by the `(f): f is Field | Tabs: ...` notation. Does this cause `field` to be Field | Tab without actually specifying the return type of find (which I expect to take an arrow function to boolean?)

frontend/src/js/external-forms/FormsNavigation.tsx
```diff
@@ -89,7 +89,6 @@ const FormsNavigation = ({ reset }: Props) => {
           value={options.find((o) => o.value === activeForm) || null}
           onChange={(value) => {
             if (value) {
-              reset();
```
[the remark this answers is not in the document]
awildturtok: You could add a comment like `we intentionally don't reset so that values are kept when switching forms`

frontend/src/js/external-forms/helper.ts
```diff
@@ -1,10 +1,27 @@
...
+} from "./config-types";
 
 const nonFormFieldTypes = new Set(["HEADLINE", "DESCRIPTION"]);
 
+// Different forms may have fields with the same name.
+// We want to remember values of fields of form A when switching to form B,
+// so users may come back to form A and see their previous values.
+// So in order to avoid field name clashes, we need unique field names
+export const getUniqueFieldname = (formType: string, field: Field | Tabs) => {
+  return `${formType}--${field.name}`;
```
awildturtok, line 19: There might be an even better place for this (where you fetch the field names), where you assert that no field contains `--`  since that would break getRawFieldName.

frontend/src/js/external-forms/stateSelectors.ts
```diff
@@ -31,22 +32,23 @@ export const selectRunningQuery = (state: StateT) => {
 };
 
 function getVisibleConceptListFields(
-  config: { fields: GeneralField[] },
+  config: Form,
+  fields: GeneralField[],
```
awildturtok, line 36: So if I understand your code right, this `fields` is global, for ALL available forms at once? If that's the case, why don't you try reworking fields to be a map `config["type"] -> GeneralField[]` (since that's what you accomplished in effect either way). Thereby avoiding having to virtualise the field-name.

awildturtok (commented):
There's nothing big, but I find the prefix approach a bit of a hack. If there's a good reason for it in your reducer architecture, go ahead but otherwise you might be better off by using a map as I outlined.

frontend/src/js/external-forms/FormConfigLoader.tsx
Kadrian: That's called a "type guard", we have quite some of those in our helper methods

frontend/src/js/external-forms/helper.ts
Kadrian: Hmm, yeah I thought about transforming directly after the fetch, not to prevent some accidental `--` within field names (I think it's a fair assumption / contract that they won't), but it would indeed avoid making the transformation at many places in the code. I'll give it a try

Kadrian:
> There's nothing big, but I find the prefix approach a bit of a hack.

It's a transformation step for the field names, not really a hack, 

> If there's a good reason for it in your reducer architecture, go ahead

To clarify: this has nothing to do with reducer architecture, but rather how `react-hook-form` keeps field state. If we want to keep field state when switching forms, all fields of all forms might be "registered" at the same time with `react-hook-form`'s FormProvider. And since they're registered with a string name, those names have to be distinct for different forms. So I don't think the prefix approach is a "hack", but I understand that it might feel like one. I'll try the "transform right after fetching" approach, maybe that makes it less "invasive".

> but otherwise you might be better off by using a map as I outlined.

I didn't really understand want you meant here

Kadrian:
@awildturtok I've given this another shot, this time performing the transformation
- right after loading forms
- right after loading a saved form config
- directly before submitting a form to the API

so now this should feel less invasive, please have another look

frontend/src/js/external-forms/form-concept-group/FormConceptCopyModal.tsx
```diff
@@ -38,25 +38,32 @@ const SxInputCheckbox = styled(InputCheckbox)`
...
   const visibleConceptListFields = useVisibleConceptListFields();
 
   const conceptListFieldOptions = visibleConceptListFields
-    .filter((field) => field.name !== targetFieldname)
+    .filter((field) => {
+      const isAnotherField = field.name !== targetFieldname;
+      const hasValues =
+        formValues[field.name] &&
+        formValues[field.name].some((value: FormConceptGroupT) =>
+          value.concepts.some(exists),
```
awildturtok, line 62: mega nitpick: but i would probably write this as `formValues[field.name].flatMap(v -> v.concepts).some(exists)`  since you're not interested in the nested context

frontend/src/js/external-forms/helper.ts
```diff
@@ -1,10 +1,27 @@
 import type { SelectOptionT } from "../api/types";
 import type { Language } from "../localization/useActiveLang";
 
-import type { FormField, GeneralField, Group } from "./config-types";
+import type {
+  Field,
+  FormField,
+  GeneralField,
+  Group,
+  Tabs,
```
awildturtok, line 9: 2 unsused imports

frontend/src/js/external-forms/helper.ts
```diff
@@ -1,10 +1,27 @@
...
+} from "./config-types";
 
 const nonFormFieldTypes = new Set(["HEADLINE", "DESCRIPTION"]);
 
+// Different forms may have fields with the same name.
+// We want to remember values of fields of form A when switching to form B,
+// so users may come back to form A and see their previous values.
+// So in order to avoid field name clashes, we need unique field names
+export const getUniqueFieldname = (formType: string, rawFieldName: string) => {
```
awildturtok, line 18: Might be a nitpick, but I think might be useful: can you make `getUniqueFieldname` return `uniqueFieldNameT` so that the string-type is tagged on the type system level or would that require you down-casting it when interacting with the form library/react? That way you're less likely to miss casts or misshandle the values. so `getUniqueFieldname : (string, string) -> uniqueFieldnameT` and `getRawFieldname: uniqueFieldnameT -> string`

and `type uniqueFieldnameT = string`

awildturtok (approved):
none of these comments are actual issues just observations, i think it looks good

frontend/src/js/external-forms/form-concept-group/FormConceptCopyModal.tsx
Kadrian: good point, done

frontend/src/js/external-forms/helper.ts
awildturtok: I did way too much googling out of curiosity, and there seems to be an approach here: https://www.typescriptlang.org/play#example/nominal-typing

It makes sense though, that this isn't a primary concern of TS (i honestly forgot it was structurally typed)

PR #3021, opened 2023-04-04
Currency Input update

fabian-bizfactory:
Draft, needs to be discussed internally

### DIFF
--- a/frontend/src/js/entity-history/timeline/EventCard.tsx
+++ b/frontend/src/js/entity-history/timeline/EventCard.tsx
@@ -155,10 +155,7 @@ const EventCard = ({
                   <TinyLabel>{column.defaultLabel}</TinyLabel>
                   <code>
                     <NumericFormat
-                      thousandSeparator={currencyConfig.thousandSeparator}
-                      decimalSeparator={currencyConfig.decimalSeparator}
-                      decimalScale={currencyConfig.decimalScale}
-                      suffix={currencyConfig.prefix}
+                      {...currencyConfig}
                       displayType="text"
                       value={parseInt(row[column.label]) / 100}
                     />
--- a/frontend/src/js/entity-history/timeline/EventCard.tsx
+++ b/frontend/src/js/entity-history/timeline/EventCard.tsx
@@ -155,9 +155,7 @@ const EventCard = ({
                   <TinyLabel>{column.defaultLabel}</TinyLabel>
                   <code>
                     <NumericFormat
-                      thousandSeparator={currencyConfig.thousandSeparator}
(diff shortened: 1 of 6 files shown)


Kadrian (changes requested):
I think the right way to do this would be
1) let's call the currency config field `unit` (instead of `suffix` or `prefix` and let's not use spaces within the string (the backend shouldn't be involved in the details on how or where the unit gets displayed in the frontend, so already calling it `prefix` wasn't very well done from our side)
2) in the frontend, where we render `NumericFormat`, let's pass 
```
suffix={" " + currencyConfig.suffix}
```
because the fact that we're using a **space** for separating the value form the unit is an implementation detail

frontend/src/js/ui-components/CurrencyInput.tsx
```diff
@@ -48,10 +48,12 @@ const CurrencyInput: FC<PropsT> = ({
...
     }
   }, [value]);
-
   return (
     <SxNumberFormat
-      {...currencyConfig}
+      thousandSeparator={currencyConfig?.thousandSeparator}
+      decimalSeparator={currencyConfig?.decimalSeparator}
+      decimalScale={currencyConfig?.decimalScale}
+      suffix={" " + currencyConfig?.unit}
```
Kadrian, line 56: Easier like this:
```suggestion
     {...currencyConfig}
      suffix={" " + currencyConfig.unit}
```

( I think we can safely make `currencyConfig` **non-optional** in `CurrencyInput` )

frontend/src/js/entity-history/timeline/EventCard.tsx
```diff
@@ -158,7 +158,7 @@ const EventCard = ({
                       thousandSeparator={currencyConfig.thousandSeparator}
                       decimalSeparator={currencyConfig.decimalSeparator}
                       decimalScale={currencyConfig.decimalScale}
-                      suffix={currencyConfig.prefix}
+                      suffix={" " + currencyConfig.unit}
```
Kadrian, line 161: Same here, we should be able to **spread** currencyConfig and override `suffix`

Kadrian (approved):
👍

awildturtok:
This PR broke things internally, please notify me if you change something in the backend's source code. The config is not just a static file, but can be provided and overriden at start time. Changing the name caused requires us to change it in our config as well.

fabian-bizfactory:
> This PR broke things internally, please notify me if you change something in the backend's source code. The config is not just a static file, but can be provided and overriden at start time. Changing the name caused requires us to change it in our config as well.

Hey, I am so sorry about that! I will double check all the PRs in the future, so that won't happen again. If there is anything needed, please let me know.

PR #3025, opened 2023-04-05
Long Result list is now scrollable

fabian-bizfactory:


Kadrian (approved):
Interesting, what was the use case? Do we really have lists that are larger than say 10 items here?

fabian-bizfactory:
> Interesting, what was the use case? Do we really have lists that are larger than say 10 items here?

The use case are long outputs (please see internal for a description). It seems like >10 would be possible.

PR #3038, opened 2023-04-18
Form editor prevent opening, when cannot drop + add highlighting

fabian-bizfactory:
This PR also unifies the code to decide if a concept is addable to the concept.

### DIFF
--- a/frontend/src/js/model/node.ts
+++ b/frontend/src/js/model/node.ts
@@ -128,7 +128,7 @@ export function getNodeIcon(
   return config?.isStructNode ? faFolderRegular : faFolder;
 }
 
-export const droppableObjectIsConceptTreeNode = (
+const droppableObjectIsConceptTreeNode = (
   node: PossibleDroppableObject,
 ): node is DragItemConceptTreeNode => {
   return node.type === DNDType.CONCEPT_TREE_NODE;
--- a/frontend/src/js/model/node.ts
+++ b/frontend/src/js/model/node.ts
@@ -15,6 +15,7 @@ import type {
   DragItemConceptTreeNode,
   StandardQueryNodeT,
 } from "../standard-query-editor/types";
+import { PossibleDroppableObject } from "../ui-components/Dropzone";
 
 import { objectHasNonDefaultSelects } from "./select";
 import {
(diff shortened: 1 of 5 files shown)


frontend/src/js/query-node-editor/ConceptDropzone.tsx
```diff
@@ -17,23 +21,35 @@ interface PropsT {
...
 }
 
+export const droppableObjectIsConceptTreeNode = (
+  node: PossibleDroppableObject,
+): node is DragItemConceptTreeNode => {
+  return node.type === DNDType.CONCEPT_TREE_NODE;
+};
+
+export const canDropConceptTreeNodeBeDropped = (node: StandardQueryNodeT) => {
+  return (item: PossibleDroppableObject) => {
```
Kadrian, line 31: Good idea to share the code. Could you simplify this a little?

- rename to `canNodeBeDropped`
- move the code to `src/js/model/node.ts`
- maybe even just use two parameters `(node, item)` instead of returning a function and inline the function where it's called?

frontend/src/js/small-tab-navigation/HoverNavigatable.tsx
```diff
@@ -28,7 +28,6 @@ const Root = styled("div")<{
     return "inherit";
   }};
   position: relative;
-  border-radius: ${({ theme }) => theme.borderRadius};
```
Kadrian, line 31: Why this change? This makes the background bleed out the container at least for the editor

frontend/src/js/query-node-editor/ConceptDropzone.tsx
fabian-bizfactory: > maybe even just use two parameters (node, item) instead of returning a function and inline the function where it's called?
I thought that it might be a little simpler , but with two params looks better so changed it to that

frontend/src/js/model/node.ts
```diff
@@ -126,3 +127,25 @@ export function getNodeIcon(
 
   return config?.isStructNode ? faFolderRegular : faFolder;
 }
+
+export const droppableObjectIsConceptTreeNode = (
```
Kadrian, line 131: no need to export this, or am I missing something?

frontend/src/js/standard-query-editor/QueryNode.tsx
```diff
@@ -197,18 +187,7 @@ const QueryNode = ({
...
-        ) {
-          return false;
-        }
-        const conceptId = item.ids[0];
-        const itemAlreadyInNode = node.ids.includes(conceptId);
-        const itemHasConceptRoot = item.tree === node.tree;
-        return itemHasConceptRoot && !itemAlreadyInNode;
-      }}
+      canDrop={(item) => canNodeBeDropped(node, item)}
       highlightDroppable={true}
```
Kadrian, line 191:
```suggestion
      highlightDroppable
```

Kadrian (approved):

PR #3041, opened 2023-04-18
Hide Copy Button Attribute

fabian-bizfactory:


frontend/src/js/external-forms/stateSelectors.ts
```diff
@@ -80,6 +80,11 @@ export const useAllowExtendedCopying = (
   // Need to have min 2 fields to copy from one to another
   if (otherConceptListFields.length < 1) return false;
 
+  const targetField = visibleConceptListFields.find(
+    (field) => field.name === targetFieldname,
+  );
+  if (targetField?.showCopyButton === false) return false;
```
Kadrian, line 86: The problem here is that when I'm configuring form fields and I'm **not** defining `showCopyButton` in a field (`showCopyButton === undefined`) would be equal to defining `showCopyButton: true`. Could be hard to understand that if I have two fields
```
{
  name: "xyz",
  showCopyButton: true
}

{
  name: "abc",
}
```

I might be led to think that for the second field, the copy button is **not** shown. But like this it would be shown in both cases.

The default behavior (= leaving out the optional attribute) should be the negative of the verb ("show" / "hide"), so:

- If by default we want to show the copy button, the optional attribute should be called `hideCopyButton`, and we should check `if (targetField?.hideCopyButton) return false`.
- If by default we want to hide the copy button, the optional attribute should be called `showCopyButton` and we should check `if (!targetField?.showCopyButton) return false` (at the top of this function)

frontend/src/js/external-forms/stateSelectors.ts
```diff
@@ -80,6 +80,11 @@ export const useAllowExtendedCopying = (
   // Need to have min 2 fields to copy from one to another
   if (otherConceptListFields.length < 1) return false;
 
+  const targetField = visibleConceptListFields.find(
+    (field) => field.name === targetFieldname,
+  );
+  if (targetField?.hideCopyButton === true) return false;
```
Kadrian, line 86:
```suggestion
  if (targetField?.hideCopyButton) return false;
```

Kadrian (approved):

PR #3045, opened 2023-04-19
add datepicker for date inputs

MarcoKorinth:


frontend/src/js/ui-components/InputDate.tsx
```diff
@@ -0,0 +1,114 @@
...
+              props.inputProps?.onKeyPress?.(e);
+            },
+          }}
+        />
+        <ReactDatePicker
+          ref={datePickerRef}
+          selected={value ? parseDate(value, dateFormat) : new Date()}
+          onChange={(val) => {
+            const selectedDate = formatDate(val as Date, dateFormat);
+            val && onChange(selectedDate);
```
Kadrian, line 100: This confused me a bit: can `val` ever be null / undefined? If yes, `formatDate` might crash and we should replace the type cast with a null check. If no, we won't need the `val &&` check. Or am I missing something else here?

frontend/src/js/ui-components/InputDate.tsx
```diff
@@ -0,0 +1,114 @@
...
+    background: ${({ theme }) => theme.col.green};
+  }
+`;
+
+type Props = Omit<BaseInputProps, "inputType"> & {
+  value: string | null;
+  dateFormat: string;
+  className?: string;
+  onChange: (val: string | null) => void;
+  onCalendarSelect?: (val: string | null) => void;
```
Kadrian, line 44: How about?
```suggestion
  onCalendarSelect?: (val: string) => void;
```

frontend/src/js/ui-components/InputDate.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+`;
+
+const HiddenInput = styled("input")`
+  display: none;
+`;
+
+const StyledCalendar = styled("div")`
+  .react-datepicker__day--selected {
+    background: ${({ theme }) => theme.col.green};
```
Kadrian, line 35: Due to legacy our primary color is called `blueGrayDark`, haha! How about using it here:
```suggestion
    background: ${({ theme }) => theme.col.blueGrayDark};
    font-weight: 700;
```

Kadrian (approved):
Really well done, I like how you've contained everything in a new `InputDate` component, plus the interaction feels natural to me 👍 

I left a couple of small comments, would be happy if you could take a look before merging. 

Thinking about the use case: most people will want "to pick whole month ranges, quarters or years", so I guess we'll have to do a follow up to see what else the library provides here (looking at some of the examples with year / month / quarter pickers & dropdowns https://reactdatepicker.com/#example-year-select-dropdown)

frontend/src/js/ui-components/InputDate.tsx
MarcoKorinth: Yes the onChange method passes a "Date | null" parameter.
I don't know in what case null is passed but to be sure I implemented a null check.

PR #3055, opened 2023-05-02
fixed datepicker autofocus issue

MarcoKorinth:
- Datepicker now opens by clicking the input field even when already focused.
- Datepicker now automatically opens if input field is autofocused

### DIFF
--- a/frontend/src/js/ui-components/InputDate.tsx
+++ b/frontend/src/js/ui-components/InputDate.tsx
@@ -1,5 +1,11 @@
 import styled from "@emotion/styled";
-import { createElement, createRef, forwardRef, useState } from "react";
+import {
+  createElement,
+  createRef,
+  forwardRef,
+  useEffect,
+  useState,
+} from "react";
 import ReactDatePicker from "react-datepicker";
 import "react-datepicker/dist/react-datepicker.css";
 
@@ -53,12 +59,21 @@ const InputDate = forwardRef<HTMLInputElement, Props>(
       onChange,
       onCalendarSelect,
       onFocus,
+      onBlur,
+      onClick,
       ...props
     },
     ref,
   ) => {
     const datePickerRef = createRef<ReactDatePicker>();
+    const [hasFocus, setHasFocus] = useState(false);
     const [focusBlocked, setFocusBlocked] = useState(false);
+
+    useEffect(() => {
+      if (props.inputProps?.autoFocus && hasFocus) {
+        datePickerRef.current?.setOpen(true);
+      }
+    }, [props.inputProps?.autoFocus, hasFocus, datePickerRef]);
 
     return (
       <Root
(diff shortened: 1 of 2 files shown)


frontend/src/js/ui-components/InputDate.tsx
```diff
@@ -53,13 +59,22 @@ const InputDate = forwardRef<HTMLInputElement, Props>(
       onChange,
       onCalendarSelect,
       onFocus,
+      onBlur,
+      onClick,
       ...props
     },
     ref,
   ) => {
     const datePickerRef = createRef<ReactDatePicker>();
```
Kadrian, line 68: Just noticing this now, we should `useRef` to avoid recreating the ref on every render
```suggestion
    const datePickerRef = useRef<ReactDatePicker>();
```

frontend/src/js/ui-components/InputDate.tsx
```diff
@@ -53,13 +59,22 @@ const InputDate = forwardRef<HTMLInputElement, Props>(
...
+    const [hasFocus, setHasFocus] = useState(false);
     const [focusBlocked, setFocusBlocked] = useState(false);
 
+    useEffect(() => {
+      if (props.inputProps?.autoFocus && hasFocus) {
+        datePickerRef.current?.setOpen(true);
+      }
+    }, [props.inputProps?.autoFocus, hasFocus, datePickerRef]);
```
Kadrian, line 76: (sidenote: no need for any refs in dependency arrays)

frontend/src/js/ui-components/InputDate.tsx
```diff
@@ -81,9 +96,18 @@ const InputDate = forwardRef<HTMLInputElement, Props>(
...
               datePickerRef.current?.setOpen(true);
             }
           }}
+          onBlur={(e) => {
+            onBlur?.(e);
+            setHasFocus(false);
+          }}
+          onClick={(e) => {
+            onClick?.(e);
+            hasFocus && datePickerRef.current?.setOpen(true);
```
Kadrian, line 109:
```suggestion
            if (hasFocus) {
              datePickerRef.current?.setOpen(true);
            }
```

frontend/src/js/ui-components/InputDate.tsx
```diff
@@ -53,13 +59,22 @@ const InputDate = forwardRef<HTMLInputElement, Props>(
...
       ...props
     },
     ref,
   ) => {
     const datePickerRef = createRef<ReactDatePicker>();
+    const [hasFocus, setHasFocus] = useState(false);
     const [focusBlocked, setFocusBlocked] = useState(false);
 
+    useEffect(() => {
+      if (props.inputProps?.autoFocus && hasFocus) {
```
Kadrian, line 73: Hmm, I think while `onClick` of the input should open the picker, we shouldn't auto-open on `autoFocus`, because that feels like "a modal the user didn't trigger". So can we maybe live without the useEffect?

Kadrian (changes requested):
Let's remove the auto-open on auto-focus

Kadrian (approved):
very nice, works well!

PR #3076, opened 2023-05-22
Feature/allow insertion of concepts between each other

fabian-bizfactory:


frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [height, setHeight] = useState(40);
+
+  const [{ isOver }, addZoneRef] = useDrop({
+    accept: acceptedDropTypes,
+    drop: onDrop,
+    hover(item) {
+      if (item.type === "CONCEPT_TREE_NODE") {
+        return setHeight(item.dragContext.height);
+      }
+      return setHeight(0);
```
Kadrian, line 46: No need to set the height if we re-order while dragging

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,63 @@
...
+  acceptedDropTypes: string[];
+}
+
+const Root = styled("div")<{
+  height: number;
+}>`
+  width: 100%;
+  left: 0;
+  top: -15px;
+  height: ${({ height }) => height + 40}px;
```
Kadrian, line 18: 40 is a magic number, let's extract variables for cases like this

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,63 @@
...
+  right: 0;
+  position: relative;
+  border-radius: ${({ theme }) => theme.borderRadius};
+`;
+
+const Expander = styled("div")<{
+  height: number;
+}>`
+  overflow: hidden;
+  margin-top: ${({ height }) => -height}px;
```
Kadrian, line 28: Let's avoid setting very dynamic numeric values in styled components (to avoid recalculating classnames) and let's use inline styles instead

frontend/src/js/external-forms/form-components/DropzoneList.tsx
```diff
@@ -87,22 +97,30 @@ const DropzoneList = <DroppableObject extends PossibleDroppableObject>(
...
-              <StyledIconButton icon={faTimes} onClick={() => onDelete(i)} />
-              {item}
-            </ListItem>
+            <div key={i}>
+              {!disallowMultipleColumns && (
+                <DropzoneBetweenElements
+                  acceptedDropTypes={acceptedDropTypes}
+                  onDrop={dropBetween(i)}
+                />
+              )}
```
Kadrian, line 106: I'd suggest to add one below the last element as well

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -237,130 +282,146 @@ const FormConceptGroup = (props: Props) => {
...
-                            j,
-                            !concept.includeSubnodes,
-                            newValue,
-                          ),
-                        ),
-                      expandable:
-                        !props.disallowMultipleColumns &&
-                        hasConceptChildren(concept),
-                      active: !!concept.includeSubnodes,
+          <>
```
Kadrian, line 285: Is this needed?

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -196,9 +203,43 @@ const FormConceptGroup = (props: Props) => {
...
             : props.attributeDropzoneText
         }
+        dropBetween={(i: number) => {
+          return (item: DragItemConceptTreeNode) => {
+            if (props.isValidConcept && !props.isValidConcept(item))
+              return null;
+
+            if (isMovedObject(item)) {
+              if (exists(item.dragContext.deleteInOrigin)) {
+                item.dragContext.deleteInOrigin();
```
Kadrian, line 216: So right now you're
- passing the delete logic in through the drag context
- running the delete logic with the side effect that the values in the form context will have updated
- re-getting the values from the form context
- inserting the value

That's a little complicated. Wouldn't it be possible to just use `props.value` right here, delete the concept or value right here and insert right here as well? So no change to `dragContext`, no call to `getValues`?

Kadrian (changes requested):
Alright, left some further comments in our issue tracking system, but here's some first code feedback as well

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
fabian-bizfactory: I agree with you that it is complicated. That were my concerns with this solution as well. However I do not see a different one for the following reason:
If someone drags between two different DropzoneLists then we can not simply delete the element in the target List. The only other solution I can imagine right now (maybe you have a better one in mind), would be to pass the field name where the dragged element originated from. The downside of that would be, that it would increase the code complexity in the deletion logic. 

On the other hand if you want to reduce the complexity and limit the functionality in the sense, that moving between two lists is no longer possible. That could be implemented as well. But I think that should be an internal discussion with the product owners.

fabian-bizfactory:
@Kadrian I updated the User Experience of the insertion. Could you take a look at let me know what you think?

Kadrian:
The UX feels a lot better to me already. How about removing any "vertical jumping" of the concepts when dragging and the drop-line shows up? So only "show / hide" the drop line? And I think the line could be slightly thicker.

I'll have a look at the code again later

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
Kadrian: Hmm. Thought about this again. What I'm trying to avoid is calling `getValues` on the own fieldname. We're passing down `props.value` for the field value. So we should avoid getting the same data from context with getValues. And I'm trying to avoid the implicit side effect of the deletion. 

I'd propose:
- to rename both deleteInForm and deleteInOrigin to be the same, e.g. `deleteFromOtherField` to signal that that's a feature that's ONLY there to support drag-drop moving **across** fields.
- to add a `movedFromFieldName` to the dragContext
- if (isMovedObject(item)), to check if the `dragContext.movedFromFieldName` is the same as the own props.fieldName. If yes, we'd do the delete + insert logic right here, starting from `props.value`, no call to `getValues` anymore. If no, we'd "call & forget" `deleteFromOtherField`, since it won't influence our own props.value. Side effect, yes, but it won't influence our logic here. Then we'd do only the insert logic here.

Does that make sense? Con:
- a little more verbose than your solution right now
- the delete logic will be "in two places"
Pro: 
- it will model more explicitly that "moving from other fields" is a feature we want
- it will avoid getting the same state from props AND context
- we won't have to deal with a possible, implicit side effect

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,70 @@
...
+      isOver: monitor.isOver(),
+      isDroppable: monitor.canDrop(),
+    }),
+  });
+
+  const rootHeightMultiplier = lastElement ? 0.5 : 1;
+  const rootDefaultMarginTop = (lastElement ? -15 : -5) - LineHeight;
+  const rootOverMarginTop = lastElement ? -23 : -10;
+  const rootDefaultTop = lastElement ? -5 : -10;
+  const rootOverTop = (lastElement ? -2 : -15) - LineHeight;
```
Kadrian, line 50: Those calculations seem a bit arbitrary with a lot of magic numbers. Are they necessary? Why does the height and marginTop need to change? Aren't the lines visually always the same height?

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,70 @@
...
+const Line = styled("div")`
+  overflow: hidden;
+  display: block;
+  background-color: ${({ theme }) => theme.col.blueGrayLight};
+  margin: 1px 0;
+  height: ${LineHeight}px;
+  border-radius: 2px;
+`;
+
+const BetweenElements = <DroppableObject extends PossibleDroppableObject>({
```
Kadrian, line 32: Let's please name exported components like the file names

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -199,6 +203,58 @@ const FormConceptGroup = (props: Props) => {
...
+                  addConcept(
+                    insertValue(updatedValue, insertIndex, newValue),
+                    insertIndex,
+                    copyConcept(item),
+                  ),
+                );
+              } else {
+                if (exists(item.dragContext.deleteFromOtherField)) {
+                  item.dragContext.deleteFromOtherField();
+                }
```
Kadrian, line 237: Could you simplify this code (and the other / similar part) to make it more readable? Maybe by destructuring the variables from dragContext?

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
fabian-bizfactory: The height needs to change so that the elements below do not bounce up and down. This affects the margin as well => we need to take away less space than before. 

I was able to take one of the magic values out and simplify some more. I think the other magic values are needed.

fabian-bizfactory: I don't think `absolute` would help us in this case, since we would need to find the position for the root by a lot more complex calculations (e.g. scroll positions etc.). I think `relative`is the intended display property.

Kadrian: When you say `absolute`, are you referring to `position: fixed`? Fixed would mean related to the viewport / scroll positions - like you say, but `absolute` simply relates to the parent container that would have to be `relative` (within DropzoneList)

fabian-bizfactory: Sorry, you are right. That would work. I will try out if that works better.

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,66 @@
...
+  top,
+}: Props<DroppableObject>) => {
+  const [{ isOver }, addZoneRef] = useDrop({
+    accept: acceptedDropTypes,
+    drop: onDrop,
+    collect: (monitor) => ({
+      isOver: monitor.isOver(),
+      isDroppable: monitor.canDrop(),
+    }),
+  });
```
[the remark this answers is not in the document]
fabian-bizfactory: I changed the Root element to a `Dropzone`.  That works as well. I am not completley convinced on updating the isOver property in the parent component. However, I could not find a better solution. Do you have any ideas on that, or do you think its fine like that?

I chose to only provide top and height for now instead of a classname, since that would cover all cases for now.

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <SxDropzone
+        bare
+        naked
+        acceptedDropTypes={acceptedDropTypes}
+        onDrop={onDrop}
+        setIsOver={setIsOver}
+        height={height}
+        top={top}
+      />
+    </>
```
Kadrian, line 55: There shouldn't be a need to add the `isOver` state outside of the component. If you render `Line` as a child of the Dropzone, you should be able to use the `isOver` prop from the children render function that `Dropzone` supplies.

Could you give that a try?

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,59 @@
...
+} from "../../ui-components/Dropzone";
+
+interface Props {
+  onDrop: (props: PossibleDroppableObject, monitor: DropTargetMonitor) => void;
+  acceptedDropTypes: string[];
+  top: number;
+  height: number;
+}
+
+const LineHeight = 3;
```
Kadrian, line 16: Now with all the changes, there's no need anymore to have this be an external variable, right?

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const Line = styled("div")<{ show: boolean }>`
+  overflow: hidden;
+  display: block;
```
Kadrian, line 20: What's the need for `display: block` and `overflow: hidden`? I think we should be able to go without those

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
```diff
@@ -0,0 +1,59 @@
...
+  visibility: ${({ show }) => (show ? "visible" : "hidden")};
+  background-color: ${({ theme }) => theme.col.blueGrayDark};
+  margin: 1px 0;
+  height: ${LineHeight}px;
```
Kadrian, line 24: (we can make this as big as `5px` I think)

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -199,6 +208,50 @@ const FormConceptGroup = (props: Props) => {
...
+              ? copyConcept(item)
+              : initializeConcept(item, defaults, tableConfig);
+            let newPropsValue = props.value;
+            let insertIndex = i;
+            if (isMovedObject(item)) {
+              const { movedFromFieldName, movedFromAndIdx, movedFromOrIdx } =
+                item.dragContext;
+
+              if (movedFromFieldName === props.fieldName) {
+                if (i > movedFromAndIdx && movedFromOrIdx === 0) {
```
Kadrian, line 228: Say I have
```
andidx 0 - concept a
andidx 1 - concept b, concept c
andidx 2 - concept d
andidx 3 - concept e
```
and I drag concept `b` down between concepts `d` and `e`, this will decrease the insertIndex and actually insert before concept `d`. That's a bug, right? What is this if condition testing exactly? Maybe it helps to extract this into a variable name?

frontend/src/js/external-forms/form-components/DropzoneBetweenElements.tsx
fabian-bizfactory: I think thats a little about personal preference. I think 5 px is a little fat for the line -> I would prefer 3px then 4px. I will push it with 4 let me know what you think

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
fabian-bizfactory: The index of where we want to insert may change if we insert below (the concept is the only one in the andidx). => Will extract and fix this in the next commit

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -396,9 +453,30 @@ const FormConceptGroup = (props: Props) => {
...
+              if (movedFromFieldName === props.fieldName) {
+                updatedValue =
+                  updatedValue[movedFromAndIdx].concepts.length === 1
+                    ? removeValue(updatedValue, movedFromAndIdx)
+                    : removeConcept(
+                        updatedValue,
+                        movedFromAndIdx,
+                        movedFromOrIdx,
+                      );
+                setEditedFormQueryNodePosition({ valueIdx, conceptIdx });
```
Kadrian, line 471: Could you explain this part – why is it required to set a potentially different node as the one to be edited?

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -199,6 +208,53 @@ const FormConceptGroup = (props: Props) => {
             ? t("externalForms.common.concept.copying")
             : props.attributeDropzoneText
         }
+        dropBetween={(i: number) => {
+          return (item: PossibleDroppableObject) => {
+            if (item.type !== DNDType.CONCEPT_TREE_NODE) return;
+
+            if (props.isValidConcept && !props.isValidConcept(item))
+              return null;
```
Kadrian, line 216: Feels inconsistent to return undefined above and null below, could you make this consistent?

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -199,6 +208,53 @@ const FormConceptGroup = (props: Props) => {
...
+              ? copyConcept(item)
+              : initializeConcept(item, defaults, tableConfig);
+            let newPropsValue = props.value;
+            let insertIndex = i;
+            if (isMovedObject(item)) {
+              const { movedFromFieldName, movedFromAndIdx, movedFromOrIdx } =
+                item.dragContext;
+
+              if (movedFromFieldName === props.fieldName) {
+                const willConceptMoveUp =
```
Kadrian, line 228: Shouldn't it be called `willConceptMoveDown`? The dragged concept is moving down, no?

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
fabian-bizfactory: When we delete a concept which is above the currently selected concept (in the Concept editing menu) then the index will change -> on rerender a different concept will be shown in the modal.

=> I extracted a variable and added a clarifying comment in the code

Kadrian: Aah I see, thanks for clarifying

PR #3091, opened 2023-06-13
Support concept column stratified infos

Kadrian:


frontend/src/js/entity-history/timeline/YearHead.tsx
```diff
@@ -102,20 +124,62 @@ const TimeStratifiedInfos = ({
...
+                const column = info.columns.find((c) => c.label === label);
+
+                if (!column) {
+                  return null;
+                }
+
+                if (isConceptColumn(column)) {
+                  const semantic = column.semantics.find(
+                    (s): s is ColumnDescriptionSemanticConceptColumn =>
+                      s.type === "CONCEPT_COLUMN",
```
awildturtok, line 136: probably would've wanted to use isConceptColumn?

Kadrian: No, here I'm not selecting a column that's a concept column (== that contains at least one semantic of type CONCEPT_COLUMN), but instead I'm selecting the CONCEPT_COLUMN semantic from within the concept column

PR #3109, opened 2023-06-28
Support default time aggregation settings

Kadrian:


frontend/src/js/concept-trees/ConceptTreeNode.tsx
```diff
@@ -121,6 +88,10 @@ const ConceptTreeNode: FC<PropsT> = ({
             matchingEntities: data.matchingEntities,
             dateRange: data.dateRange,
 
+            excludeTimestamps:
+              root.excludeFromTimeAggregation ||
```
Kadrian, line 92: to also "pass down" a date exclusion setting from the root to the concepts cc @awildturtok

PR #3135, opened 2023-07-24
Remove error context handling

Kadrian:


frontend/src/js/api/types.ts
```diff
@@ -428,9 +428,8 @@ export interface GetQueryErrorResponseT {
 }
 
 export interface ErrorResponseT {
-  code: string; // To translate to localized messages
-  message?: string; // For developers / debugging only
-  context?: Record<string, string>; // More information to maybe display in translated messages
+  message?: string; // Localized error message (based on Accept-Language header) to show to users
```
Kadrian, line 431: @awildturtok will `message` always be defined?

frontend/src/js/query-runner/actions.ts
```diff
@@ -179,12 +178,7 @@ const getQueryErrorMessage = ({
     return t("queryRunner.queryCanceled");
   }
 
-  return (
-    (error &&
-      error.code &&
-      getExternalSupportedErrorMessage(t, error.code, error.context)) ||
-    t("queryRunner.queryFailed")
-  );
+  return error?.message || t("queryRunner.queryFailed");
```
Kadrian, line 181: That's the main change in the logic: now trying to use `message` if it's there, otherwise use a generic query failed message

frontend/src/js/api/types.ts
awildturtok: @MaxVonKnobloch ?

Mine are all defined, so maybe we should enforce that also in ExternalForms

awildturtok (approved):

PR #3144, opened 2023-07-28
Fix SQL tests

torbsto:
The SQL tests are broken after recent changes to develop. This removes the condition that SQL tests are only run on branches starting with `sql/` because it seems to me like it would be better to have them run for all changes now to avoid such situations.

Additionally, there are the following fixes:

1) Namespace CastClassException 

ManagedExecution is used by both connectors. We therefore cannot cast to `DistributedNamespace` here
because this will lead to a `CastClassException` when using the SQL connector. The cast was introduced in b4bc03774b1a9778c9023382f26f17eab82de184. For the resulting exception, see the following CI run: https://github.com/ingef/conquery/actions/runs/5655732144/job/15321399831.

This pushes the cast down to `ManagedInternalForm`. We could also think about introducing an additional parent class for all ManagedExecution classes except `SqlManagedQuery`, in which the cast is done.

2) I18n Initalization

Because the error messages use the C10n interface (changed in 52a3b9fce91ee9fb0340b90bad0a52fe76829c23), the error message in one of our tests was wrong. This changes the expected error messageand adds the initialization for I18n

### DIFF
--- a/.github/workflows/test_backend.yml
+++ b/.github/workflows/test_backend.yml
@@ -41,5 +41,4 @@ jobs:
       - name: JSON based Integration Tests
         run: mvn test -T 1C -pl backend -Dgroups="INTEGRATION_JSON"
       - name: SQL based Integration Tests
-        if: ${{ startsWith(github.head_ref, 'sql/') }}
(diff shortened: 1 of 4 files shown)


jnsrnhld (approved):
LGTM.

PR #3152, opened 2023-08-16
Fix Highlighting in Tooltip

fabian-bizfactory:


Kadrian:
I like that you went for an own approach, tackling the Typescript compatibilities, without adding more libraries for highlighting.

But now the code is hard to read, contains a lot of type casts and edge case handling, uses `Children` methods (which is discouraged https://react.dev/reference/react/Children#alternatives ) and has to handle all those different `components` as part of react-markdown. 

Could you maybe give this another try with the following idea:
- create or find a remark plugin that traverses the markdown tree and wraps text matches into highlighted `span`s
- use only `span` in `components` to match these highlighted spans and then render a `<HighlightedText>`

Kadrian (changes requested):
See comment

fabian-bizfactory:
I have just looked over the list of plugins again and found one that supports marks in Markdown `==`.

I think that will be the easiest solution.

frontend/src/js/tooltip/Tooltip.tsx
```diff
@@ -158,6 +159,11 @@ const ConceptLabel = ({
   );
 };
 
+const mark = (text: string, words: string[]): string => {
+  const regex = new RegExp(words.join("|"), "gi");
```
Kadrian, line 163: Guess I would have used a `useMemo` to avoid calling `new RegExp` again and again when the `infos` change frequently on hovering around, while the `words` stay the same. But I quickly checked the performance impact and couldn't notice much. Hmm!

PR #3164, opened 2023-08-31
fix: datepicker blocks input field

MarcoKorinth:
# Changes
- fixed bug: datepicker opens over input field
- datepicker only opens after clicking the new calendar icon inside the input field
- datepicker starts in year selection mode by default


frontend/src/js/common/helpers/mergeRefs.ts
```diff
@@ -0,0 +1,17 @@
+type Mutable<T> = {
+  -readonly [k in keyof T]: T[k];
+};
```
Kadrian, line 3: Hmmm in other implementations I can see them using a `MutableRefObject` type from React. You probably copied the code below from somewhere, right? I guess I wonder whether we shouldn't just use some external implementation like https://github.com/gregberge/react-merge-refs but up to you

Kadrian (changes requested):
I'd be happy if you tried reusing `IconButton`, otherwise looks pretty good!

Kadrian (approved):
Nice!
Slightly funny that the year picker half way overlays the day picker if there aren't that many years. But I hope that will just be fixed in the backend with more available years, cc @awildturtok 

Here's a small suggestion https://github.com/ingef/conquery/pull/3166

awildturtok:
@MarcoKorinth kannst du bitte die dependencies auf eva seite gleichstellen?

PR #3176, opened 2023-09-19
Keep row prefix constant when reordering

fabian-bizfactory:


Kadrian (changes requested):
Hmm, what you'd be doing here is "hard-coding" the name `row_prefix`. So `rowPrefixField` is an optional select field with a dynamic name, and in this case it might be `row_prefix`, but in other fields in might be another name, so we'll have to rely on `field.rowPrefixField.name` for the prefix field name.

Guess what I'd try instead would be to
- pass down a `rowPrefixFieldname` and maybe the `row` itself from the FormConceptGroup to the FormConceptNode
- add both to the dragContext
- get them both from the dragContext in dropBetween
- if defined, get the prefix field value from the row + rowPrefixFieldname and pass a modified `newValue` to `insertValue()`

Does that make sense?

fabian-bizfactory:
> Hmm, what you'd be doing here is "hard-coding" the name `row_prefix`. So `rowPrefixField` is an optional select field with a dynamic name, and in this case it might be `row_prefix`, but in other fields in might be another name, so we'll have to rely on `field.rowPrefixField.name` for the prefix field name.
> 
> Guess what I'd try instead would be to
> 
>     * pass down a `rowPrefixFieldname` and maybe the `row` itself from the FormConceptGroup to the FormConceptNode
> 
>     * add both to the dragContext
> 
>     * get them both from the dragContext in dropBetween
> 
>     * if defined, get the prefix field value from the row + rowPrefixFieldname and pass a modified `newValue` to `insertValue()`
> 
> 
> Does that make sense?

I just implemented some changes to this PR. `rowPrefixFieldname` is only passed. I don't think the `row` needs to be passed `props.value[movedFromAndIdx]` returns the `row` as well. 

Please let me know what you think

Kadrian:
Would that work with concepts dragged in from other fields as well?

fabian-bizfactory:
No, but that wasn't possible in the previous version either. If we would allow that we would need to take in the possible values for the prefix into account, wouldn't we? To catch the case of the prefixes having different possible values.

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -231,14 +231,39 @@ const FormConceptGroup = (props: Props) => {
...
-                newPropsValue =
+                let newPropsValue =
                   props.value[movedFromAndIdx].concepts.length === 1
                     ? removeValue(props.value, movedFromAndIdx)
                     : removeConcept(
                         props.value,
                         movedFromAndIdx,
                         movedFromOrIdx,
                       );
+                // Having a row prefix is a special property that is only used in an edge case form. We only need to pass it
```
awildturtok, line 242:
```suggestion
                // Having a row prefix is a special property that is only used in an edge case form, used for tagging concepts. We only need to pass it
```

frontend/src/js/external-forms/form-concept-group/FormConceptGroup.tsx
```diff
@@ -231,14 +231,39 @@ const FormConceptGroup = (props: Props) => {
...
+                // back into the value if the concept is moved to a different position in the same field
+                const modifiedValue = newValue as unknown as {
+                  [index: string]: string;
+                };
+                if (
```
awildturtok, line 247: puuh, das sieht gut so aus, aber kP ob das bei dir Nebeneffekte hat, die nicht offensichtlich sind.

zB darf newValue mutable behandelt werden?

fabian-bizfactory: Danke, da hast du Recht, dass prop sollte so eigentlich nicht geändert werden. Ich habe gerade eine Version gepusht die ein deep copy umsetzt (mit json)

Kadrian (approved):
@fabian-bizfactory I've cleaned up some of the logic, check my last commit. Was a bit easier than trying to explain. Looks good otherwise.

PR #3178, opened 2023-09-21
feat: renew datepicker year selection

MarcoKorinth:
# What changed
- The year selection for datepickers now span over 100 years.
- The layout for the year selection changed from grid layout to simple 1 column list

### DIFF
--- a/frontend/src/js/ui-components/InputDate/CustomHeader.tsx
+++ b/frontend/src/js/ui-components/InputDate/CustomHeader.tsx
@@ -57,10 +57,11 @@ const SelectMenu = ({
   onSelect,
 }: Pick<ReactDatePickerCustomHeaderProps, "date"> & {
   options: SelectOptionT[];
-  layout: "grid" | "flex";
+  layout: "oneColumn" | "twoColumns";
   onSelect: (n: number) => void;
 }) => {
-  const OptionList = layout === "grid" ? MonthListContainer : YearListContainer;
+  const OptionList =
+    layout === "twoColumns" ? MonthListContainer : YearListContainer;
   return (
     <SelectMenuContainer>
       <Menu>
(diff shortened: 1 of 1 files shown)


frontend/src/js/ui-components/InputDate/CustomHeader.tsx
```diff
@@ -27,12 +25,20 @@ export const SelectMenuContainer = styled("div")`
...
   display: grid;
   grid-template-columns: auto auto;
   gap: 5px;
 `;
 
+export const YearListContainer = styled(List)`
+  display: flex;
+  flex-direction: column-reverse;
+  gap: 5px;
+  height: 200px;
```
Kadrian, line 38: Is that the picker height? Maybe extract this as a variable?

frontend/src/js/ui-components/InputDate/CustomHeader.tsx
```diff
@@ -46,32 +52,37 @@ export const MonthYearLabel = styled("div")`
 
 const SelectMenu = ({
   date,
+  layout,
   options,
   onSelect,
 }: Pick<ReactDatePickerCustomHeaderProps, "date"> & {
   options: SelectOptionT[];
+  layout: "grid" | "flex";
```
Kadrian, line 60: That prop confused me a little, since it's not really grid vs flex but "one column" vs "two columns", guess I'd just pass the respective OptionList component itself as a prop, but yeah, just a "nit".

Kadrian (approved):
Looks good to me!

frontend/src/js/ui-components/InputDate/CustomHeader.tsx
MarcoKorinth: No, it's just the height of the year selection container.
Unfortunately the height of the datepicker is dynamic and can't be extracted that easily.

frontend/src/js/ui-components/InputDate/CustomHeader.tsx
MarcoKorinth: Yes you're right, I renamed the layout options

PR #3218, opened 2023-11-16
wraps realm in abstract binder

thoniTUB:
Removes the following warning, wehen the ApiTokenRealm was configured:
```log
WARN  [2023-11-15 14:52:08]     o.g.j.i.i.Providers             A provider com.bakdata.conquery.models.auth.apitoken.ApiTokenRealm registered in SERVER runtime does not implement any provider interfaces applicable in the SERVER runtime. Due to constraint configuration problems the provider com.bakdata.conquery.models.auth.apitoken.ApiTokenRealm will be ignored.
```

However, the ApiTokenRealm was still usable

### DIFF
--- a/backend/src/main/java/com/bakdata/conquery/models/config/auth/ApiTokenRealmFactory.java
+++ b/backend/src/main/java/com/bakdata/conquery/models/config/auth/ApiTokenRealmFactory.java
@@ -23,6 +23,7 @@ import lombok.Data;
 import org.apache.http.util.CharArrayBuffer;
 import org.apache.shiro.authc.AuthenticationToken;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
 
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/bakdata/conquery/models/config/auth/ApiTokenRealmFactory.java
```diff
@@ -45,8 +46,13 @@ public ConqueryAuthenticationRealm createRealm(ManagerNode managerNode) {
 		managerNode.getAuthController().getAuthenticationFilter().registerTokenExtractor(new ApiTokenExtractor());
 
 		JerseyEnvironment environment = managerNode.getEnvironment().jersey();
-		environment.register(apiTokenRealm);
 
+		environment.register(new AbstractBinder() {
```
awildturtok, line 50: lol hä?

awildturtok (approved):

PR #3219, opened 2023-11-16
throws RTE on emtpy test sets for programmatic test

thoniTUB:
Tiny proposal so unintentional skipped tests sets are more obvious.


awildturtok (approved):
Danke!

PR #3238, opened 2023-12-04
shrink in single thread

awildturtok:


backend/src/test/java/com/bakdata/conquery/integration/sql/dialect/HanaSqlIntegrationTests.java
```diff
@@ -48,7 +48,7 @@
 @Slf4j
 public class HanaSqlIntegrationTests extends IntegrationTests {
 
-	private final static DockerImageName HANA_IMAGE = DockerImageName.parse("saplabs/hanaexpress:2.00.061.00.20220519.1");
+	private final static DockerImageName HANA_IMAGE = DockerImageName.parse("saplabs/hanaexpress:2.00.072.00.20231123.1");
```
thoniTUB, line 51: @jnsrnhld @awildturtok I had to update the tag here, because HANA EXPRESS seems to have only one active tag at a time. This will happen from time to time again and is anoying 😢

thoniTUB (approved):

PR #3310, opened 2024-02-22
remove apitoken related classes and code

thoniTUB:
This removes a shiro realm that allowed to create API-Tokens on behave of a user, which have a subset of the Users Capabilities.

The realm is removed because:
- It was not used
- There are other mechanics to achieve the same goal but are compatible with other authentication mechanism, such as creating a service account with limited rights
- The hashing was flawed, as it used a static hash


awildturtok (approved):

thoniTUB:
@awildturtok habe das target auf release gesetzt, weil der PR untergegangen war

PR #3337, opened 2024-03-07
Feature/clean up ftl

thoniTUB:
defines a ftl marco that containes the head section and the surrounding html, which is now used also by the unauthenticated resources (login and availiable logins).

Makes availiable logins look like the local login:  
image

### DIFF
--- a/backend/src/main/resources/com/bakdata/conquery/resources/admin/ui/templates/README.md
+++ b/backend/src/main/resources/com/bakdata/conquery/resources/admin/ui/templates/README.md
@@ -1,0 +1,4 @@
+# Debuging FTL
+
+Sometimes FTL errors are printed nowhere, e.g. if an import fails.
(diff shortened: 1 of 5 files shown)


backend/src/main/resources/com/bakdata/conquery/resources/admin/ui/templates/README.md
```diff
@@ -0,0 +1,4 @@
+# Debuging FTL
+
+Sometimes FTL errors are printed nowhere, e.g. if an import fails.
+In this case but a breakpoint in `freemarker.template.TemplateExceptionHandler#DEBUG_HANDLER`.
```
thoniTUB, line 4:
```suggestion
In this case put a breakpoint in `freemarker.template.TemplateExceptionHandler#DEBUG_HANDLER`.
```

awildturtok (approved):

PR #3349, opened 2024-03-18
Move frontend Dockerfile to root; add git describe

Kadrian:


thoniTUB:
Looks goot to me.
I tried to identify other files that need to be updated, but all deployments work on the produced images from the github workflow and don't build the image themselves.

@awildturtok do you see any problems regarding eva?

thoniTUB (approved):

PR #3358, opened 2024-03-21
Simplify Table rendering. Move allocations out of Cell-functions, avoiding unnecessary allocations

awildturtok:
Die Methoden auszulagern fehlt mir die Zeit aktuell, würde den Code aber deutlich lesbarer machen.
Ich werde demnächst einen PR stellen, indem ich das mit der currency-Code richtig implementiere, dann ist das auch sauberer.

### DIFF
--- a/frontend/src/js/preview/tableUtils.ts
+++ b/frontend/src/js/preview/tableUtils.ts
@@ -4,9 +4,7 @@ import { useTranslation } from "react-i18next";
 import { useSelector } from "react-redux";
 import { CurrencyConfigT, GetQueryResponseDoneT } from "../api/types";
 import { StateT } from "../app/reducers";
-import {
-  NUMBER_TYPES, currencyFromSymbol
-} from "./util";
+import { NUMBER_TYPES, currencyFromSymbol } from "./util";
 
 export type CellValue = string | Vector;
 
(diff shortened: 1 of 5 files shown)


frontend/src/js/preview/tableUtils.ts
```diff
@@ -20,7 +15,18 @@ export function useCustomTableRenderers(queryData: GetQueryResponseDoneT) {
...
+      const dateFormatter = new Intl.DateTimeFormat(navigator.language, {
+        day: "2-digit",
+        month: "2-digit",
+        year: "numeric",
+      });
+
+      const currencyFormatter = new Intl.NumberFormat(navigator.language, {
+        style: "currency",
+        currency: currencyFromSymbol(currencyConfig.unit),
+      });
```
Kadrian, line 28: How about extracting these out, so we don't re-initialize the formatters on every cell render?

awildturtok: The changes in Table.tsx#72 should ensure, that this method is called only once per column.

Kadrian: Aah, I see 👍

frontend/src/js/preview/tableUtils.ts
```diff
@@ -20,7 +15,18 @@ export function useCustomTableRenderers(queryData: GetQueryResponseDoneT) {
   );
 
   const getRenderFunction = useCallback(
-    (cellType: string): ((value: CellValue) => string) | undefined => {
+    (cellType: string): ((value: CellValue) => string) => {
+      const dateFormatter = new Intl.DateTimeFormat(navigator.language, {
```
Kadrian, line 19: Good idea actually to use this new API instead of our usual `dateHelper` functions.

Kadrian (approved):
Looks good to me! 👍 Good idea to limit to 100 lines for now and to avoid manually attaching window scroll listeners and doing scroll pos calculations

PR #3392, opened 2024-04-12
adds and removes headers for index.html to prevent it from being hashed

thoniTUB:
```
* Rebuilt URL to: localhost:7777/
* Uses proxy env variable no_proxy == 'localhost,
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 7777 (#0)
> GET / HTTP/1.1
> Host: localhost:7777
> User-Agent: curl/7.61.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: nginx/1.24.0
< Date: Fri, 12 Apr 2024 06:37:55 GMT
< Content-Type: text/html
< Content-Length: 992
< Last-Modified: Fri, 12 Apr 2024 06:37:29 GMT
< Connection: keep-alive
< Vary: Accept-Encoding
< Cache-Control: no-store
< Accept-Ranges: bytes
<
<!DOCTYPE html>
<html lang="en">
```
ETag removed and Cache-Control: no-store.

Other assets remain cachable, such as js and css


frontend/container/etc/nginx/conf.d/default.conf
```diff
@@ -0,0 +1,30 @@
+server {
+    listen       80;
+
+    gzip         on;
+    gzip_vary    on;
+    gzip_types   application/javascript text/css;
```
Kadrian, line 6: Is there a reason to restrict this to those types?

frontend/container/etc/nginx/conf.d/default.conf
```diff
@@ -0,0 +1,30 @@
...
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+	
+	location ~ ^/(index\.html)?$ {
+            add_header Cache-Control "no-store";
+            etag off;
+	}
+    }
+
+    #error_page  404              /404.html;
```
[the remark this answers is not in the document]
Kadrian: yeah, let's remove all commented out code

frontend/container/etc/nginx/conf.d/default.conf
```diff
@@ -0,0 +1,30 @@
...
+
+    gzip         on;
+    gzip_vary    on;
+    gzip_types   application/javascript text/css;
+
+    #access_log  /var/log/nginx/host.access.log  main;
+
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
```
Kadrian, line 12: There is no index.htm

frontend/container/etc/nginx/conf.d/default.conf
thoniTUB: Good point, should be fine to zip every thing if it is performant using the wildcard. text/html seems to be always compressed.

PR #3438, opened 2024-05-14
Handle internal server error on upload query

Kadrian:


frontend/src/js/previous-queries/upload/UploadQueryResults.tsx
```diff
@@ -59,18 +58,20 @@ const UploadQueryResults = ({
...
         setUploadResult(e as UploadQueryResponseT);
       } else {
         dispatch(
           setMessage({
             message: t("uploadQueryResultsModal.uploadFailed"),
-            type: SnackMessageType.ERROR,
+            type: "error",
           }),
         );
       }
```
Kadrian, line 77: @awildturtok so if I understand this correctly, we seem to allow 400s as valid responses. But I received an internal server error that was still using a 400 status code, not a 500. So I've added a `"resolved" in e` check, which is far from ideal. Hmm.

Screenshot 2024-05-14 at 11 53 46

thoniTUB: Do I understand it correctly, that you would expect a 200ish response if nothing could be uploaded?

Kadrian: Yes, together with the following response shape
```
export type UploadQueryResponseT = {
  resolved: number;
  unresolvedId: string[][];
  unreadableDate: string[][];
};
```

and in case it's an internal server error, I'd expect a 500, not a 4xx.

PR #3469, opened 2024-06-14
Fix/separate manager restart

thoniTUB:
Removes the shutdown message, as we don't want to shut down shards, when running distributed, and when we run standalone, the shards are getting already managed by the environment.


backend/src/main/java/com/bakdata/conquery/mode/cluster/ClusterConnectionManager.java
```diff
@@ -108,8 +102,6 @@ public void start() throws IOException {
 	}
 
 	public void stop() {
-		clusterState.getShardNodes().forEach(((socketAddress, shardNodeInformation) -> shardNodeInformation.send(new ShutdownShard())));
```
awildturtok, line 111: hat eh nie geklappt :D

awildturtok (approved):

PR #3508, opened 2024-07-30
fixes Column-Select creation of TableExportQueryConverter. Current im…

awildturtok:
…pl has issues with selection of SecondaryId-Columns


awildturtok:
@jnsrnhld der kann weg oder?

jnsrnhld:
Kann weg

PR #3536, opened 2024-08-27
remove cqpp import via file path

thoniTUB:
One solution to fix 3 of these warnings: https://github.com/ingef/conquery/security/code-scanning

An alternative would be to define a path (e.g.) in PreprocessingConfig that specifies an absolute path from which local CQPPs are allowed to be imported.


awildturtok (approved):
Ne löschen ist hier deutlich sinnvoller. Danke

PR #3582, opened 2024-10-02
Fix search for multi struct node hierarchy

Kadrian:
So that all nested levels of struct nodes open, if one of the concept trees contained within them is part of the search result.


Kadrian:
Wrong base branch

PR #3643, opened 2024-12-18
upgrade to JKD21. Also upgrades groovy deps, as those would no longer build.

awildturtok:


thoniTUB (commented):
Dockerfile bitte noch anpassen und testen

backend/pom.xml
```diff
@@ -163,10 +163,10 @@
...
         </dependency>
         <dependency>
-            <groupId>org.codehaus.groovy</groupId>
-            <artifactId>groovy</artifactId>
-            <version>3.0.13</version>
-            <classifier>indy</classifier>
+            <groupId>org.apache.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
+            <version>4.0.24</version>
+<!--            <classifier>indy</classifier>-->
```
thoniTUB, line 169:
```suggestion
```

PR #3658, opened 2025-01-28
split initialization of matching stats and filter indices into separa…

awildturtok:
…te endpoints:

Since matching stats are now serialized, we now only need to re-initialize searches on startup


backend/src/test/java/com/bakdata/conquery/integration/tests/ConceptResolutionTest.java
```diff
@@ -41,13 +41,7 @@ public void execute(StandaloneSupport conquery) throws Exception {
...
 		test.importRequiredData(conquery);
 
-		final URI matchingStatsUri = HierarchyHelper.hierarchicalPath(conquery.defaultAdminURIBuilder()
-															, AdminDatasetResource.class, "postprocessNamespace")
-													.buildFromMap(Map.of(DATASET, conquery.getDataset().getId()));
-
-		conquery.getClient().target(matchingStatsUri)
-				.request(MediaType.APPLICATION_JSON_TYPE)
-				.post(null);
+		LoadingUtil.initializeIndices(conquery);
```
thoniTUB, line 44: Ist das für diesen Test eigentlich relevant?

thoniTUB (approved):
👏

PR #3679, opened 2025-02-25
Draft for failing testcase

jnsrnhld:


awildturtok:
@jnsrnhld ich hatte einen copy paste Fehler. Konnte es jetzt reproduzieren. Hoffentlich hilft das in der Eingrenzung!

backend/src/main/java/com/bakdata/conquery/sql/conversion/model/select/UniversalSqlSelect.java
jnsrnhld, line 1: Just a confusing class which had no benefit.

backend/src/test/java/com/bakdata/conquery/integration/sql/CsvTableImporter.java
```diff
@@ -75,6 +74,10 @@ public void importTableIntoDatabase(RequiredTable requiredTable) {
 	}
 
 	private void insertValuesIntoTable(Table<Record> table, List<Field<?>> columns, List<RowN> content, Statement statement) throws SQLException {
+		// encountered empty new line
```
jnsrnhld, line 77: There was a second bug you uncovered with your example, because the CSV importer was importing empty lines into the table as a all-`null` values. This caused results like: 
```csv
result,dates,concept connector_1 exists
,{},
1,{},1
```

backend/src/main/java/com/bakdata/conquery/sql/conversion/model/select/ExistsSqlSelect.java
```diff
@@ -2,21 +2,52 @@
...
 import java.util.List;
-
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
 import org.jooq.Field;
+import org.jooq.Name;
 import org.jooq.impl.DSL;
 
-public class ExistsSqlSelect extends UniversalSqlSelect<Integer> {
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
```
jnsrnhld, line 11: To keep a reference onto selects from the beginning of the conversion on to the last step, we usually create an `ExtractingSelect`. But this disables the ability to know what kind of `Select` the reference originated from, because it's not more than a qualified name. 

By overriding this behavior for the `ExistsSelect`, we can keep this knowledge and aggregate it in later steps.

awildturtok (approved):
Ich werde den Test nochmal vervollständigen. Danke für den Fix!

PR #3700, opened 2025-03-17
Fix: sync on the synchronized set

thoniTUB:
I guess we got the synchronization wrong here.

The constructor uses itself as the barrier not the delegate:
```
        SynchronizedCollection(Collection<E> c) {
            this.c = Objects.requireNonNull(c);
            mutex = this;
        }
```

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -114,7 +114,7 @@
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-to-slf4j</artifactId>
-            <version>2.24.0</version>
+            <version>2.24.3</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/bakdata/conquery/io/storage/xodus/stores/CachedStore.java
```diff
@@ -126,11 +127,8 @@ public void loadKeys() {
 
 	@Override
 	public Stream<KEY> getAllKeys() {
-		// We need to synchronize around this and mutations to avoid ConcurrentModificationExceptions.
-		synchronized (backingKeys) {
-			// This creates a humongous copy, but avoids concurrent modification and simplifies usage.
-			return Set.copyOf(keys).stream();
-		}
+		// Uses .toArray() under the hood, which is synchronized
+		return new ArrayList<>(keys).stream();
```
thoniTUB, line 131:
```suggestion
		return (Stream<KEY>) Arrays.stream(keys.toArray());
```

PR #3767, opened 2025-07-16
Fix: Race condition on loadStorage

thoniTUB:


backend/src/main/java/com/bakdata/conquery/io/storage/xodus/stores/CachedStore.java
```diff
@@ -153,14 +154,15 @@ public void loadData() {
 
 		final Stopwatch timer = Stopwatch.createStarted();
 
-		final Set<KEY> dupes = new HashSet<>();
+		final Object VALUE_DUMMY = new Object();
+		final ConcurrentHashMap<KEY,Object> dupes = new ConcurrentHashMap<>();
```
awildturtok, line 158: Collections.newSetFromMap(new ConcurrentHashMap<>())

awildturtok (dismissed):

PR #3873, opened 2026-04-08
feat: use unprivileged nginx image for frontend

thoniTUB:
- Uses nginx-unprivileged as base container image
- Uses templating to set container port NGINX is listening on using the environment variable NGINX_PORT (default port: **8000**).
  Because the port is set upon startup, the container does not EXPOSE a port by default anymore
- integrated the environment replacement script in the docker entry point instead of the container command

### DIFF
--- a/frontend.Dockerfile
+++ b/frontend.Dockerfile
@@ -1,5 +1,5 @@
 # Version Extractor
-FROM alpine/git:2.52.0 AS version-extractor
+FROM alpine/git:2.52.0@sha256:d453f54c83320412aa89c391b076930bd8569bc1012285e8c68ce2d4435826a3 AS version-extractor
 
 WORKDIR /app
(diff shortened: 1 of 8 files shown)


frontend.Dockerfile
```diff
@@ -1,13 +1,13 @@
 # Version Extractor
-FROM alpine/git:2.52.0 AS version-extractor
+FROM alpine/git:2.52.0@sha256:d453f54c83320412aa89c391b076930bd8569bc1012285e8c68ce2d4435826a3 AS version-extractor
```
awildturtok, line 2: warum hier eine SHA?

thoniTUB: Best practise für Reproduzierbarkeit :)

awildturtok (approved):

PR #3885, opened 2026-05-05
Feature/ch support

awildturtok:


awildturtok:
pulled into https://github.com/ingef/conquery/pull/3886 because they needed the same fixes

PR #3921, opened 2026-07-16
Merge EventFilterCte into PreprocessingCte

awildturtok:
EventFilterCte was a leftover from a more modular setup, but was causing issues with performance due to relabeling losing access to indices.


awildturtok:
Superseded  by https://github.com/ingef/conquery/pull/3928/changes

PR #3957, opened 2026-08-26
chore(frontend): migrate to pnpm 11 / node 24, replace eslint+prettier with biome

Kadrian:
Two tooling-only commits, no runtime behaviour change.

**pnpm 11 + node 24** (node 18 is EOL, pnpm 11 needs >=22)
- lockfiles converted via `pnpm import`, resolved versions unchanged (except @types/node 24, tsx)
- strict node_modules exposed undeclared imports (history, react-router, tippy.js, @storybook/manager-api) - fixed instead of enabling hoisting
- `pnpm-workspace.yaml`: supply-chain policies (7d minimumReleaseAge, trustPolicy no-downgrade, build-script allowlist)
- ts-node/esm loader crashes on node >=22, mock-api/serve now use tsx
- `frontend.Dockerfile.dockerignore`: build context is repo root so `frontend/.dockerignore` never applied, host node_modules got copied into the image
- tsconfig lib es2022 (code already used `.at()`; @types/node 18 had masked it)

**biome instead of eslint + prettier**
- `biome.jsonc` keeps parity with the old eslint rules; recommended-preset rules that would need real code changes are off/warn with inline reasons - follow-up PR
- scripts: `pnpm check` / `pnpm fix`, CI runs `biome ci`
- source diff = biome's safe fixes (`import type`, import order, `catch {}`, `Object.hasOwn`) + a few one-line fixes for genuine errors

Follow-ups: private-side mirror PR (submodule bump), dependency upgrades as separate PR.


awildturtok (approved):
Top, dass du auch direkt an eva gedacht hast

frontend/biome.jsonc
thoniTUB, line 1: @Kadrian hier sind noch ein paar TODOs drinne. Willst du die noch mitnehmen oder sind die out of scope für den PR?

frontend/pnpm-workspace.yaml
```diff
@@ -0,0 +1,11 @@
+# Supply-chain hardening (see https://pnpm.io/settings)
+# Refuse versions published less than 7 days ago
+minimumReleaseAge: 10080
```
thoniTUB, line 3: This might need coordination with dependabots cooldown feature

frontend/biome.jsonc
Kadrian: out of scope --> follow-up, otherwise this won't be reviewable,

frontend/pnpm-workspace.yaml
Kadrian: good point. We'll get to this

thoniTUB (approved):

PR #3961, opened 2026-08-26
feat: adds prioritized list of ID-kinds

awildturtok:
Mapped from IdColumnConfig


backend/src/main/java/com/bakdata/conquery/apiv1/frontend/FrontendPreviewConfig.java
```diff
@@ -15,16 +18,23 @@ public static class Labelled {
 		private final String name;
 		private final String label;
 	}
-
-	private final Collection<Labelled> all;
+  
+  private final Collection<Labelled> all;
```
thoniTUB: space indent

thoniTUB (approved):

PR #3983, opened 2026-08-27
chore(frontend): tailwind-variants, theme tokens, global styles in css

Kadrian:
First step of the emotion → tailwind migration.

- `tailwind-variants` in, `tailwind-styled-components` out; its 7 usages are `tv()` now. No `tailwind-merge`: tv 3 ships its own class merger, `className` overrides go through it (`heading3({ className })`); pnpm still resolves it as an optional peer in the lockfile, not a direct dep
- conventions: every class list lives in a `tv()`, conditionals are variants, longer lists use the array form with one utility per line (only tightly coupled ones share a line, e.g. `absolute top-3 right-4`, `opacity-80 hover:opacity-100`)
- `@theme` gains what only the emotion theme had: `red`/`green`/`orange`, `filetype-*`, `text-tiny` (11px), bare `rounded` = 3px. Other font sizes map onto tailwind defaults (xs/sm/base/xl/2xl)
- `GlobalStyles.tsx` → `@layer base` in `index.css`, tippy overrides too; dead `.SplitPane`/`.Resizer` rules dropped
- tab headline / snack close are `<button>`s now (a11y rules see raw elements, `tw()` hid them)

Follow-ups: migrate `styled()` per directory (~200 files), then drop `@emotion/*` + shrink `conquery({ theme })` to `{ img, palette }`. Downstream repo must mirror the dependency change and already overrides the tokens via `:root`.


frontend/src/js/pane/TabNavigation.tsx
```diff
@@ -14,29 +14,16 @@ const Root = styled("div")`
...
-    $active ? "text-primary-500" : "text-gray-500 hover:text-black"};
-  ${({ $active }) =>
-    $active
-      ? "border-primary-500"
-      : "border-transparent hover:border-primary-200"};
-
-`;
+const headline = tv({
+  base: "mt-[6px] mr-[5px] mb-0 shrink-0 cursor-pointer border-b-[3px] px-3 text-sm font-bold uppercase leading-[30px] tracking-wider transition-colors",
+  variants: {
```
awildturtok, line 31: cool!

thoniTUB (approved):

awildturtok (approved):

PR #4006, opened 2026-09-03
refactor(frontend): Button with intent and size replaces the text buttons

Kadrian:
**Design-system prep, step 6: the Button**

Stacked on #4005. `ui-components/Button` replaces `PrimaryButton`, `TransparentButton`, `DestroyButton` and `SelectFileButton`: one component, two axes, fixed heights. `IconButton` follows in the next step.

```tsx
<Button intent="primary" size="md" onPress={save}><Icon icon={faCheck} />Save</Button>
<Button intent="tertiary" aria-label="Delete"><Icon icon={faTrash} /></Button>
```

- `intent`: `primary` (filled), `secondary` (outlined, the default), `tertiary` (ghost with hover background), `danger` (outlined red, fills on hover), `link` (reads as a text link, no box, no padding; the dropzones' import buttons)
- `size`: `sm` 24, `md` 30 (the default, 88 of the app's 119 buttons), `lg` 36 px, each a fixed height with its own text size and icon gap, so text buttons and icon buttons line up; a button whose only children are icons is a square of the same height
- pressed state is `aria-pressed`, which colors the button; the `active` flag is gone
- react-aria's API throughout: `onPress` (press events don't bubble, the one `stopPropagation` went), `isDisabled`, `autoFocus`, `excludeFromTabOrder`
- `Button` stories: intents, all sizes with text, icon plus text and icon-only, pressed state


awildturtok (approved):
Danke auch für die Tests!

PR #4028, opened 2026-09-17
refactor(frontend): multi select on react-aria's multiple-selection combobox

Kadrian:
**Fields on react-aria, step 2: the multi select**

`ComboBoxMultiField` replaces `InputMultiSelect` on react-aria's ComboBox in multiple-selection mode; the single `ComboBoxField` shares its parts.

- the chosen values are tags in the frame (react-aria TagGroup), the list opens in a portaled popover and stays open while picking; a picked option stays listed and a second pick removes it; Backspace in the empty input removes the last tag
- both comboboxes focus the first option while typing, so Enter picks it, or creates the typed text where `creatable`
- server search: page one reloads on typing and on open, the next page on scroll through react-aria's load-more item; insert all appends on both paths and fetches the bulk in pages of 500, the backend's cap, instead of one request that failed above it
- suggestions are used as the backend sends them: the Mustache mapping had been dead since `templateValues` went away; `mustache` and `downshift` leave the dependencies
- the mock api's autocomplete answers `{ values, total }` paged like the real one; stories under `FormComponents/ComboBoxMultiField`, one simulating a server search
- seven call sites re-pointed, unlabeled ones named with new i18n keys; the preview's picker keeps its own input style


awildturtok (approved):

Kadrian:
I'll have to make some minor tweaks here next week

PR #4033, opened 2026-09-17
feat(frontend): news popover in the top bar

Kadrian:
**News popover in the top bar**

(_Next, we'll definitely have to get typography more consistent throughout the app_)

Stacked on #4032. Tells users what changed: a megaphone button next to the header menu opens a scrollable list of news items, newest first, each with its date.

- Items come per dataset from `GET /api/datasets/:id/news` as `{ id, title, description, link, date }`; the endpoint doesn't exist in the backend yet, so this is built against the mock api and the URL is a placeholder
- Without items (or without the endpoint) the button doesn't render
- Unread items carry a dot, so does the button; an item turns read once its end stayed in view for a moment
- Read ids live in the browser's user settings, per user name; items older than 60 days never count as unread
- New `UnreadDot` ui-component with a story
- Mock api: `versions` and help links in the frontend config, the app crashed against the mock without `versions`

Screenshot 2026-09-17 at 17 49 42

Screenshot 2026-09-17 at 17 52 00


frontend/src/js/news/useNews.ts
```diff
@@ -0,0 +1,98 @@
...
+const useNewsItems = () => {
+  const datasetId = useDatasetId();
+  const getNews = useGetNews();
+  const [items, setItems] = useState<NewsItemT[]>([]);
+
+  useEffect(() => {
+    let isCurrent = true;
+
+    async function loadNews() {
+      if (!datasetId) return;
```
awildturtok, line 34: hier fehlt denke ich auch ein setItems([]) um die news zu clearen?

